PR #61, opened 2018-11-29
fix: carbon styling of the <body> tag, and prevent duplicate CSS

### LINKED ISSUE #60: Review use of core carbon scss with regards to duplicates and missing styles.
## Brief
Some core styles are missing (e.g. bx--body) and when using the library styles are duplicated.

## Detailed description
Currently, neither the storybook or users receive some core styling such as bx--body and plex fonts. This is because each component is importing only the styles specific to that component, missing out a number of settings and base styling.

Further to this when this library is used in another project style definitions can be duplicated. This is most obvious with bx--label which is used in a number of components. While the core SCSS libraries in carbon-components do put in place strategies to prevent this, they rely on a globally compiled file. However, .vue files appear to be  treated as individual sandboxes which are then concatenated post-compile in Vue-CLI.

The strategy of importing only component specific files to ensure a user only gets the styles for the components they use may, as a result, be flawed.

While a Vue-CLI 3 project does by deafult remove duplicates in production, it does not in development which causes dev tools to slow down significantly.

## Review and document strategy for carbon-components-vue and usage.

moores2:
Closes #60 

fix: carbon styling of the <body> tag, and prevent duplicate CSS

#### Changelog

**New**

Add carbon styles to <body> so that it can be inherited by components.
Prevent single file components generating duplicate CSS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ storybook-static
 
 # Config files
 .npmrc
+.nvmrc
 
 # Cache folders
 .cache
--- a/README.md
+++ b/README.md
@@ -79,6 +79,11 @@ View available Vue.js components [here](http://vue.carbondesignsystem.com). Usag
 
 ## Usage
 
+### General
+
+The components do not import any of the carbon styles themselves. Use the SCSS or CSS from carbon-components to provide the styling.
+You can also use the unpkg cdn to bring in the styles wholesale - unpkg.com/carbon-components/css/carbon-components.css aliases the latest css file.
+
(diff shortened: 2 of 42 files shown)


asudoh:
I have seen this issue when several React Carbon apps used CSS-in-JS technique. I saw the biggest one is reset styles, where we rely on "import-once" technique to eliminate the duplicates. Setting `$css--reset: false` will alleviate that. I haven't had a chance to investigate further details, and thus I see the change in this PR would be most reasonable for now.

Issue #217, opened 2019-02-18
Migrate to carbon X


lee-chase:
Carbon styling is in the process of being updated, there are some breaking changes in the HTML structure.

DO NOTE BREAK V1.

Steps to migrate

- [x] Create an experimental branch with styling switch
- [x] Document components with breaking changes and create issues.

Components with non-breaking changes
- [x] CvAccording
- [x] CvButton
- [x] CvCodeSnippet
- [x] CvContentSwitcher
- [x] CvDatePicker
- [x] CvDropdown

Pending in PR #272 
- [x] CvBreadcrumb
- [x] CvCheckbox
- [x] CvDataTAble
- [x] CvFileUploader
- [x] CvForm
- [x] CvInlineLoader
- [x] CvInlineNotification
- [x] CvLink
- [x] CvList
- [x] CvLoading
- [x] CvModal
- [x] CvNumberInput
- [x] CvOverflowMenu
- [x] CvPagination
- [x] CvRadioButton
- [x] CvSkeletonText
- [x] CvSlider
- [x] CvTabs
- [x] CvToastNotification
- [x] CvTag
- [x] CvTextInput
- [x] CvTextArea
- [x] CvTile
- [x] CvTimePicker
- [x] CvToggle
- [x] CvToolbar

Not started
- [x] CvProgress
- [x] CvSearch
- [x] CvSelect
- [x] CvStructuredList

lee-chase:
NOTE: If there are significant or tricky DOM changes create a c10 file as per dropdown and date picker. Otherwise copy accordion-item changes.

lee-chase:
Moved v9 components to 10

Issue #230, opened 2019-03-04
Refactor repo for multiple packages (monorepo)


dcwarwick:
We have data-viz components inside this repo, even though they really ought to build into a separate npm package from the core components. Also, we have hit a few problems recently with builds and Babel versions, which stem from the fact that we are trying to run all our builds/storybook-builds/tests/lints/formats/etc from a single top-level package.json and dependency tree. Also, our dependency list has got a bit heavy because the dependencies have to cover everything required for core and dataviz (and, for dev dependencies, everything also required for storybook and tests etc).

We could move the data-viz components out into a separate repo, as react-data-viz has done. However, this becomes quite clunky for updating in synch, and is not very scalable if we get many more sub-packages.

A modern approach to these challenges is to structure the repo as a *monorepo* -- one repo divided into multiple packages each with its own package configuration, build and test, and dependencies. We think this approach makes sense for carbon-components-vue.

This will involve the following refactoring steps:

* add _lerna_ to the repo, and configure for individual sub-package versions.
* create `packages\common` as a private lerna sub-package containing reused elements. This will have its own `package.json`, build and test tasks, and `src` folder containing the source files.
* create `packages\core` as a public lerna sub-package named '@carbon/vue' containing the core components. This will have its own `package.json`, build and test tasks, and `src` folder containing the source files.
* create `packages\dataviz` as a public lerna sub-package named '@carbon/vue-data-viz' containing the data-viz components. This will have its own `package.json`, build and test tasks, and `src` folder containing the source files.
* create `storybook` as a private lerna sub-package containing the storybook configuration. This will have its own `package.json` and storybook build tasks.
* configure the top-level `package.json` to enable repo-wide linting and formatting tasks, as well as launchers for multi-package tasks such as build and test.

PR #250, opened 2019-03-11
feat: carbon 10 update for dropdown

lee-chase:
Part of #217

Adds carbon 10 features to CvDropdown behind a feature flag.


lee-chase:
Do not merge until after #230 is completed.

Issue #206, opened 2019-02-13
Adding .babelrc prevents "yarn build"


lee-chase:
## Detailed description

With .babelrc in the root folder it is not possible to run "yarn build" successfully.

lee-chase:
@dcwarwick moving this file to test/unit/.babelrc seems to give me a working build and tests.

Can you confirm this works for you?

PR #275, opened 2019-03-21
Refactor as monorepo with lerna

### LINKED ISSUE #230: Refactor repo for multiple packages (monorepo)

dcwarwick:
Closes #230 #206 

Add lerna and refactor as a monorepo. Split core and data-viz source trees and builds. Separate storybook story source and builds.

#### Changelog

**New**

- lerna added

**Changed**

- source tree refactored


dcwarwick:
closing and reopening to retrigger checks

PR #394, opened 2019-05-05
Pagination fixes

mojodna:
When `numberOfItems` changes (e.g. is loaded from a remote source and available after `mounted`), this preserves the current page (or drops it to the last available page) instead of jumping to the last page.

#### Changelog

**Changed**

- Pagination: preserves current page when `numberOfItems` changes

### DIFF
--- a/packages/core/src/components/cv-pagination/cv-pagination.vue
+++ b/packages/core/src/components/cv-pagination/cv-pagination.vue
@@ -137,7 +137,7 @@ const newPagesArray = pageCount => {
   return Array.from({ length: pageCount }, (val, key) => key + 1);
 };
 
-const newFirstItem = (pageValue, pageSizeValue) => Math.max(0, 1 + (pageValue - 1) * pageSizeValue);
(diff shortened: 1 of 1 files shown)


packages/core/src/components/cv-pagination/cv-pagination.vue
```diff
@@ -137,7 +137,7 @@ const newPagesArray = pageCount => {
   return Array.from({ length: pageCount }, (val, key) => key + 1);
 };
 
-const newFirstItem = (pageValue, pageSizeValue) => 1 + (pageValue - 1) * pageSizeValue;
+const newFirstItem = (pageValue, pageSizeValue) => Math.max(0, 1 + (pageValue - 1) * pageSizeValue);
```
lee-chase: Not sure the Math.max(0,...) is needed in newFirstItem. Are you OK to remove it?

I realise it's been moved from numberOfItems() but pageValue and pageSizeValue should never be negative.

mojodna: I was ending up in a state where `pageValue = 0`, but I think that got fixed in another way. Will do.

PR #395, opened 2019-05-05
Enable searching without providing an actions slot

mojodna:
This enables searching when `@search` is present but no `actions` slot is provided.

#### Changelog

**Changed**

- Enables searching when no `actions` slot is provided

### DIFF
--- a/packages/core/src/components/cv-data-table/cv-data-table.vue
+++ b/packages/core/src/components/cv-data-table/cv-data-table.vue
@@ -34,7 +34,7 @@
           </div>
         </div>
 
-        <div v-if="$slots.actions && !batchActive" class="bx--toolbar-content">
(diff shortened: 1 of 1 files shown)


packages/core/src/components/cv-data-table/cv-data-table.vue
```diff
@@ -34,7 +34,7 @@
           </div>
         </div>
 
-        <div v-if="$slots.actions && !batchActive" class="bx--toolbar-content">
+        <div v-if="($slots.actions && !batchActive) || $listeners.search" class="bx--toolbar-content">
```
lee-chase: Search should not be available if batchActive === true

Perhaps 

(($slots.actions || $listeners.serach) && !batchActive)

mojodna: Fixed.

PR #585, opened 2019-10-09
fix(cv dropdown): inline dropdown arrow icon missing

### LINKED ISSUE #584: cv-dropdown: Inline dropdown arrow icon missing
## Detailed description

Given the dropdown is set to `inline`
After a option has been chosen from the dropdown, the caret icon disappears

Bug

`cv-dropdown`

The caret icon to the left of the text is expected to remain after a selection has been made, see https://www.carbondesignsystem.com/components/dropdown/code

Chrome Version 77.0.3865.90

Latest / Storybook

PartnerWorld/none

1. Open `cv-dropdown` in storybook http://vue.carbondesignsystem.com/?path=/story/components-cvdropdown--default
2. Enable the `inline` knob
3. Select an option in the dropdown

## Additional information

Screencap:
Screenshot 2019-10-08 at 15 51 33

mrsideshowjack:
Closes #584 

Fixes the arrow icon being replaced by the `internalValue` when set (`this.$refs.valueContent.innerHTML = child.internalContent;`) by changing the `valueContent` to the span where the text should be rather than the parent of the span and the icon.

#### Changelog

`M       packages/core/src/components/cv-dropdown/cv-dropdown.vue`

### DIFF
--- a/packages/core/src/components/cv-dropdown/cv-dropdown.vue
+++ b/packages/core/src/components/cv-dropdown/cv-dropdown.vue
@@ -50,8 +50,8 @@
         @click="onClick"
       >
         <WarningFilled16 v-if="isInvalid" class="bx--dropdown__invalid-icon" />
-        <li v-if="inline" class="bx--dropdown-text" ref="valueContent">
-          <span class="bx--dropdown-text__inner">{{ placeholder }}</span>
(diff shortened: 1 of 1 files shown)


lee-chase:
Thank you very much for spotting and fixing this @mrsideshowjack

Issue #182, opened 2019-02-07
Add some more tests


lee-chase:
Currently, only the CvTag component has any tests.

Add more tests (not all at once). Either create a work item or reference this issue.

Tests for
- [x] CvAccordion
- [x] CvBreadcrumb
- [x] CvButton
- [x] CvCheckbox
- [x] CvCodeSnippet
- [ ] CvContentSwitcher
- [ ] CvDataTable
- [ ] CvDatePicker
- [ ] CvDropdown
- [ ] CvFileUploader
- [x] CvNumberInput
- [x] CvInlineLoading
- [x] CvInlineNotification
- [x] CvLink
- [x] CvList
- [x] CvLoading
- [ ] CvModal
- [ ] CvMultiSelect
- [ ] CvOverflowMenu
- [ ] CvPagination
- [ ] CvProgress
- [x] CvRadioButton
- [x] CvSearch
- [ ] CvSelect
- [ ] CvSkeletongText
- [ ] CvSlider
- [ ] CvStructuredList
- [x] CvTabs
- [x] CvTag
- [x] CvTextArea
- [x] CvTextInput
- [ ] CvTile
- [ ] CvTimePicker
- [x] CvToastNotification
- [x] CvToggle
- [x] CvToolbar
- [x] CvTooltip
- [ ] CvUIShell
- [x] CvWrapper

## Note: Tests should include 
 - prop type, default and required checks
 - Snapshots covering each property option
 - Functional tests based on usage

ZrianinaMariia:
@lee-chase I saw `cv-button.test.js` file. Does it mean tests for CvButton are done?

ZrianinaMariia:
Also quite some test for CvButton are failing

lee-chase:
> @lee-chase I saw `cv-button.test.js` file. Does it mean tests for CvButton are done?

Thanks, fixed the button tests.

davidnixon:
I think we can close this now.  Everything in Vue 3 has a test.

PR #603, opened 2019-10-16
feat: add tests for cvCheckbox

ZrianinaMariia:
#182 182

Add some tests for CvCheckbox component

#### Changelog

A       packages/core/__tests__/cv-checkbox.test.js


ZrianinaMariia:
> Looks good. We should probably also have tests that verify:
> 
> * Changes to checked property affect the computed value isChecked (wrapper.vm.isChecked)
> * Clicking updates the value isChecked (wrapper.vm.isChecked)
> * Using modelValue property overrides checked.
> * Using an array with modelValue functions as expected.

Please let me know if there are still some tests missing.

Issue #167, opened 2019-02-04
Add skeleton components


lee-chase:
Many components have a loading form, referred to as a skeleton the Carbon Design System.

See https://www.carbondesignsystem.com/patterns/loading

Add skeleton components (not all at once). Either create a work item or reference this issue.

derick-montague:
@lee-chase - Looks like the Carbon documentation link was changed from ` https://www.carbondesignsystem.com/patterns/loading` to `https://www.carbondesignsystem.com/patterns/loading-pattern/`. The link above is directed to the 404 page.

davidnixon:
- **progress** https://v7-react.carbondesignsystem.com/?path=/story/components-progressindicator--skeleton
- **search** - https://v7-react.carbondesignsystem.com/?path=/story/components-search--skeleton
- **structured list**  https://v7-react.carbondesignsystem.com/?path=/story/components-structuredlist--skeleton
-  **text input** https://v7-react.carbondesignsystem.com/?path=/story/components-textinput--skeleton

mateusbandeiraa:
@davidnixon I'm working on implementing `CvTextInputSkeleton`. Feedback is welcome. #1542

davidnixon:
We will keep this one open. Would be great to get some more of the skeletons implemented. Especially good for hacktoberfest

PR #606, opened 2019-10-16
feat: add CvToggleSkeleton and story

mrsideshowjack:
Issue #167 

Adds `CvToggleSkeleton` component. Updates `CvToggleSkeleton` story. [Based on carbon-react's toggle skeleton implementation](https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Toggle/Toggle.Skeleton.js)

#### Changelog

```
A       packages/core/src/components/cv-toggle/cv-toggle-skeleton.vue
M       packages/core/src/components/cv-toggle/index.js
M       packages/core/src/index.js
M       storybook/stories/cv-toggle-story.js
```


packages/core/src/components/cv-toggle/cv-toggle-skeleton.vue
```diff
@@ -0,0 +1,23 @@
+<template>
+  <div class="cv-toggle bx--form-item">
+    <input disabled type="checkbox" class="bx--toggle bx--skeleton" :id="uid" />
```
lee-chase: The react version omits the following disabled attribute and the id attribute. As a result the input cannot be clicked.

An alternative if we wish to keep the id attribute is to use '@click.prevent' on the input. As a result the colors match the react component.

packages/core/src/components/cv-toggle/cv-toggle-skeleton.vue
```diff
@@ -0,0 +1,23 @@
...
+
+    <label class="bx--toggle__label bx--skeleton" :for="uid" :aria-label="label ? 'null' : label">
```
lee-chase: Allow label to be set to a default not 'null'

packages/core/src/components/cv-toggle/cv-toggle-skeleton.vue
```diff
@@ -0,0 +1,23 @@
...
+  </div>
+</template>
+
+<script>
+import uidMixin from '../../mixins/uid-mixin';
+export default {
+  name: 'CvToggleSkeleton',
+  mixins: [uidMixin],
+  props: {
+    label: 'Toggle is loading',
```
lee-chase: Should be

label: { type: String, default: 'Toggle is loading' }

PR #638, opened 2019-10-25
feat: add tests for cvnumberinput

ZrianinaMariia:
#182 

Add tests for CvNumberInput. Added default value for `label` prop for CvNumberInput component.

#### Changelog

A       packages/core/__tests__/__snapshots__/cv-number-input.test.js.snap
M       packages/core/__tests__/_helpers.js
A       packages/core/__tests__/cv-number-input.test.js
M       packages/core/src/components/cv-number-input/cv-number-input.vue

#### Additional information

I am not sure I implemented the snapshot tests correctly.
 
Test Coverage:

image

### DIFF
--- a/packages/core/__tests__/__snapshots__/cv-number-input.test.js.snap
+++ b/packages/core/__tests__/__snapshots__/cv-number-input.test.js.snap
@@ -1,0 +1,126 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`CvNumberInput should match snapshot when all props are default 1`] = `
+<cv-wrapper-stub tagtype="div" class="cv-number-input bx--form-item">
+  <div data-numberinput="" class="bx--number"><label for="1" class="bx--label"></label>
+    <!---->
+    <div class="bx--number__input-wrapper"><input id="1" type="number">
+      <!---->
+      <div class="bx--number__controls"><button type="button" aria-label="Increase number input" class="bx--number__control-btn up-icon">
+          <caretupglyph-stub></caretupglyph-stub>
+        </button> <button type="button" aria-label="Decrease number input" class="bx--number__control-btn down-icon">
+          <caretdownglyph-stub></caretdownglyph-stub>
+        </button></div>
+    </div>
+    <!---->
+  </div>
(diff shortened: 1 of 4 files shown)


ZrianinaMariia:
> Using a full mount will resolve cv-wrapper to a div or empty wrapper. However, we should test cv-wrapper separately and not test again here.

Could you please elaborate a bit? Do you mean I should remove tests related to `formItem` prop?

> Snapshots look good, need to add a slot version for helpText and invalidMessage

Added two tests `should match snapshot when helper-text slot is provided` and `should match snapshot when invalid-message slot is provided`. Let me know if more tests are needed.

> This test looks like a duplicate 'should increase value by 1 when doUp' of 'should emit Number when value prop is Number'

Done

> Value changes on down are not tested.

Done

> Cursor key up/down are not tested.

I tried to add two tests, but I could not make `trigger('keydown')` work. 

``` it('should emit value increased by 1 on Up arrow key', () => {
    const id = '1';
    const value = 2;
    const wrapper = shallow(CvNumberInput, { propsData: { id, value } });
    wrapper.find('input').trigger('keydown.up')
    expect(wrapper.emitted().input[0]).toEqual([value + 1]);
  })

  it('should emit value decreased by 1 on Down arrow key', () => {
    const id = '1';
    const value = 10;
    const wrapper = shallow(CvNumberInput, { propsData: { id, value } });
    wrapper.find('input').trigger('keydown.down')
    expect(wrapper.emitted().input[0]).toEqual([value - 1]);
  })
```

> I think we should add comments separating test sections as follows ..

Done. I saw you added sections to your PR for breadcrumb tests even if there were no corresponding tests. I did the same here.

------

Please feel free to improve the tests' names.

lee-chase:
Tests are good.

Sorry for initial ramblings, I think we need to test cv-wrapper separately, do not change the formItem test.

Testing keydown/keyup does not work because the input swallows and processes as an input. As we are relying on standard input behaviour I no longer believe we need that test.

NOTE: In order to test keydown/keyup we would have to explicitly code for it in the component and prevent default. I do not believe this would be a good idea.

PR #670, opened 2019-11-10
feat: add tests for CvLink

ZrianinaMariia:
#182 

Add tests for CvLink component.

Could not get rid of the following warning when `router-link` is used:

> [Vue warn]: Unknown custom element: <router-link> - did you register the component correctly?

I true to use `localVue` and `stubs` as described here https://vue-test-utils.vuejs.org/guides/using-with-vue-router.html, but it did not help.

#### Changelog

A       packages/core/__tests__/__snapshots__/cv-link.test.js.snap
A       packages/core/__tests__/cv-link.test.js


lee-chase:
Not sure what's happening there. Maybe if we can get to Vue-CLI 4 it would be worth raising an issue.

ZrianinaMariia:
Do you think it worth commenting the tests that involve `router-link` for now?

lee-chase:
Yes, good idea.

PR #682, opened 2019-11-14
fix: CvComboBox set input placeholder to label

### LINKED ISSUE #679: CvComboBox: placeholder/label property will not change
# Detailed description

`CvComboBox`'s label property will not change the placeholder within the element, it will remain as 'Filter'

bug 🐛

`CvComboBox`

expected the placeholder to change to text I inputted, but the placeholder remains as 'Filter'

Chrome

2.18.0

no

1. open combo box storybook (http://vue.carbondesignsystem.com/?path=/story/components-cvcombobox--default)
2. change `label` prop
3. observe the placeholder not changing

## Additional information

- Additionally it would be more consistent with multi select etc if the prop `label` was instead `placeholder` ?
- Additionally 2: The storybook 'Notes' tab for CvComboBox is showing info on `CvMultiSelect` (http://vue.carbondesignsystem.com/?path=/info/components-cvcombobox--default)

mrsideshowjack:
Closes #679 

Sets the `placeholder` prop of in `input` element to the `{{label}}` vue prop

#### Changelog

`M       packages/core/src/components/cv-combo-box/cv-combo-box.vue`

### DIFF
--- a/packages/core/src/components/cv-combo-box/cv-combo-box.vue
+++ b/packages/core/src/components/cv-combo-box/cv-combo-box.vue
@@ -47,7 +47,7 @@
           role="combobox"
           :aria-expanded="open"
           autocomplete="off"
-          placeholder="Filter"
+          placeholder="label"
           v-model="filter"
           @input="onInput"
           @focus="inputFocus"
--- a/packages/core/src/components/cv-combo-box/cv-combo-box.vue
+++ b/packages/core/src/components/cv-combo-box/cv-combo-box.vue
@@ -47,7 +47,7 @@
           role="combobox"
           :aria-expanded="open"
           autocomplete="off"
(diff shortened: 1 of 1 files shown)


lee-chase:
Thanks for taking a look at this. You need to add a colon on the front of the placeholder setting to set it to react to the change of the label property.

packages/core/src/components/cv-combo-box/cv-combo-box.vue
```diff
@@ -47,7 +47,7 @@
           role="combobox"
           :aria-expanded="open"
           autocomplete="off"
-          placeholder="Filter"
+          placeholder="label"
```
lee-chase: as per comment should be 

:placeholder="label"

to make this reactive.

PR #685, opened 2019-11-14
fix: update and emit dataValue on clearFilter

### LINKED ISSUE #684: CvComboBox: Clear filter button does not clear v-model
## Detailed description

The v-model/value does not get updated when the clear icon button is clicked by the user

Bug 🐛

`CvComboBox`

expected the v-model/value to update to and empty value when the clear icon button is clicked, but the value remained the same. As a result our interface did not reflect the update and change accordingly

chrome

2.18.0

no

1. open cvComboBox - v-model in storybook (http://vue.carbondesignsystem.com/?path=/story/components-cvcombobox--vmodel)
2. select an item in CvComboBox
3. click on the clear filter icon button and observe the v-model/value have remained the same

## Additional information

Screenshot 2019-11-14 at 15 36 42

mrsideshowjack:
Closes #684 

The v-model was not updating when the clear icon button was clicked, this PR edits the method of that button `clearFilter()` to update the internal value to an empty string `''` (which will update `this.dataValue`), and then emit this change.

#### Changelog

`M       packages/core/src/components/cv-combo-box/cv-combo-box.vue`

### DIFF
--- a/packages/core/src/components/cv-combo-box/cv-combo-box.vue
+++ b/packages/core/src/components/cv-combo-box/cv-combo-box.vue
@@ -206,10 +206,12 @@ export default {
   },
   methods: {
     clearFilter() {
+      this.internalUpdateValue('');
       this.filter = '';
       this.$refs.input.focus();
       this.doOpen(true);
       this.updateOptions();
+      this.$emit('change', this.dataValue);
     },
     checkHighlightPosition(newHiglight) {
       if (this.$refs.list && this.$refs.option) {


packages/core/src/components/cv-combo-box/cv-combo-box.vue
```diff
@@ -206,10 +206,12 @@ export default {
   },
   methods: {
     clearFilter() {
+      this.internalUpdateValue('');
       this.filter = '';
```
lee-chase: It should be possible to remove this.filter = ''; as it will be set in internalUpdateValue

mrsideshowjack: I won't get updated in `internalUpdateValue()` because in `internalUpdateValue()` it won't update `this.filter` unless it can find it in `this.dataOptions`

```
const filterOption = this.dataOptions.find(item => item.value === val);
if (filterOption) {
   this.filter = filterOption.label;
}
```

Would it be best to keep `this.filter = '';` inside `clearFilter()` and not edit the `internalUpdateValue` function?

lee-chase: Your right, thanks for checking.

Issue #671, opened 2019-11-11
Cannot use invalid-message slot for CvTextArea component


sabinead:
I am using CvTextArea to let the user enter multiple email addresses. There may be different failures for several email addresses which I want to show below the text area, f.e.

![image](https://user-images.githubusercontent.com/35919161/68573381-e5193b00-0467-11ea-98ca-8e8178a5031a.png)

I was only able to see the slot if I set the property invalid-message in addition. But then, when clearing the message, the message disappears but the text area box is still highlighted.

I have created a codepen for the 3 scenarios [https://codepen.io/sadmard/pen/xxxJrGy](https://codepen.io/sadmard/pen/xxxJrGy):
- invalid-message property: ok
- invalid-message slot: no error message
- invalid-message slot and property: highlighting is not cleared if error is cleared

Enter error in the text area to see the error, anything else should clear the error.

lee-chase:
Can you create a codepen demonstrating the issue?

This codepen seems to demonstrate what you are trying to achieve.
https://codepen.io/lee-chase/pen/qBBMdOm

sabinead:
@lee-chase 
My codepen demonstrates the issue with Modal2("Show Modal invalid-message slot") and Modal3("Show Modal invalid-message slot and property").  Modal1("Show Modal invalid-message property") works fine, but as I want to show more than one line as error I need to use the slot variant.

To reproduce the issue use the following steps in my codepen:
1. open one of the 3 modal dialogs
2. type in error
    expected: the text area is highlighted and the error message is shown
    Modal1: ok
    Modal2: no error is shown
    ![image](https://user-images.githubusercontent.com/35919161/68751781-77554680-0602-11ea-8cad-a17d3ba0a83f.png)
    Modal3: ok
    ![image](https://user-images.githubusercontent.com/35919161/68751620-30675100-0602-11ea-9204-e91e84b7c7a2.png)

2. type anything else
    expected: the text area is no longer highlighted and the error message is no longer shown
    Modal1: ok
    Modal2: no change as it didn#t show an error
    Modal3: text area is still highlighted
    ![image](https://user-images.githubusercontent.com/35919161/68751416-d36b9b00-0601-11ea-84d2-eade6250b783.png)

lee-chase:
Hmm. It would appear that some time ago I made the mistake of thinking $slots was reactive, it is not. Perhaps if the property 'invalid-message' could be truthy or string, we could decide whether or not to show the invalid-message.

[...]

PR #689, opened 2019-11-17
feat: add tests for cv-text-area component|

ZrianinaMariia:
#182

Add tests for CvTextArea component. Noticed, that isInvalid and isHelper computed values return Number, not Boolean - changed that.

#### Changelog

A       packages/core/__tests__/__snapshots__/cv-text-area.test.js.snap
A       packages/core/__tests__/cv-text-area.test.js
M       packages/core/src/components/cv-text-area/cv-text-area.vue

### DIFF
--- a/packages/core/__tests__/__snapshots__/cv-text-area.test.js.snap
+++ b/packages/core/__tests__/__snapshots__/cv-text-area.test.js.snap
@@ -1,0 +1,61 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`CvTextArea should render correctly 1`] = `
+<div class="cv-text-area bx--form-item"><label for="1" class="bx--label">test label</label>
+  <!---->
+  <div class="bx--text-area__wrapper">
+    <!----> <textarea id="1" class="bx--text-area"></textarea></div>
+  <!---->
+</div>
+`;
+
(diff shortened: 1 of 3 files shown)


lee-chase:
isInvalid and isHelper are internal truthy or falsy values which I had not bothered to convert to a boolean. If we do think it's necessary I would suggest not using a Boolean constructor, but instead two not operators !!

NOTE: Due to another issue I have refactored these computed values across 11 files. Probably best to change them all at once if we think this is important.

ZrianinaMariia:
> I would suggest not using a Boolean constructor, but instead two not operators !!

Yap.

> I would suggest not using a Boolean constructor, but instead two not operators !!

It was a bit confusing when I was testing then and the values of isInvalid and isHelper I got was length of the invalidMessage and helperText props.  I just expected these computed values to be Boolean, but I don't insist on that 😄I can adjust the tests and add a comment.

> NOTE: Due to another issue I have refactored these computed values across 11 files. Probably best to change them all at once if we think this is important.

Could you please elaborate on that? What issue?

lee-chase:
#671 highlighted that $slots are not reactive. As a result, another way to evaluate whether slots were being supplied. In each case a method called checkSlots is called on mounted and beforeUpdate.

PR #691, opened 2019-11-18
chore: make slot checks boolean

lee-chase:
Change slot checks to boolean values so they show as true/false during debug/testing.

### DIFF
--- a/packages/core/src/components/cv-data-table/cv-data-table-row.vue
+++ b/packages/core/src/components/cv-data-table/cv-data-table-row.vue
@@ -51,11 +51,11 @@ export default {
   },
   mounted() {
     // NOTE: this.$slots is not reactive so needs to be managed on beforeUpdate
-    this.dataExpandable = !!(this.$slots.expandedContent !== undefined);
+    this.dataExpandable = !!this.$slots.expandedContent;
     this.$parent.$emit('cv:mounted', this);
   },
   updated() {
-    this.dataExpandable = !!(this.$slots.expandedContent !== undefined);
+    this.dataExpandable = !!this.$slots.expandedContent;
   },
   beforeDestroy() {
     this.$parent.$emit('cv:beforeDestroy', this);
--- a/packages/core/src/components/cv-data-table/cv-data-table-row.vue
+++ b/packages/core/src/components/cv-data-table/cv-data-table-row.vue
@@ -51,11 +51,11 @@ export default {
   },
   mounted() {
     // NOTE: this.$slots is not reactive so needs to be managed on beforeUpdate
-    this.dataExpandable = this.$slots.expandedContent !== undefined;
+    this.dataExpandable = !!(this.$slots.expandedContent !== undefined);
     this.$parent.$emit('cv:mounted', this);
   },
   updated() {
-    this.dataExpandable = this.$slots.expandedContent !== undefined;
+    this.dataExpandable = !!(this.$slots.expandedContent !== undefined);
   },
   beforeDestroy() {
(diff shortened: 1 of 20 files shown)


packages/core/src/components/cv-data-table/cv-data-table-row.vue
```diff
@@ -51,11 +51,11 @@ export default {
   },
   mounted() {
     // NOTE: this.$slots is not reactive so needs to be managed on beforeUpdate
-    this.dataExpandable = this.$slots.expandedContent !== undefined;
+    this.dataExpandable = !!(this.$slots.expandedContent !== undefined);
```
ZrianinaMariia: I guess here it is already Boolean, so no need for `!!` or?

packages/core/src/components/cv-data-table/cv-data-table-row.vue
```diff
@@ -51,11 +51,11 @@ export default {
...
     this.$parent.$emit('cv:mounted', this);
   },
   updated() {
-    this.dataExpandable = this.$slots.expandedContent !== undefined;
+    this.dataExpandable = !!(this.$slots.expandedContent !== undefined);
```
ZrianinaMariia: I guess here it is already Boolean, so no need for !! or?

packages/core/src/components/cv-ui-shell/cv-header.vue
```diff
@@ -31,10 +31,10 @@ export default {
   },
   mounted() {
     // NOTE: this.$slots is not reactive so needs to be managed on beforeUpdate
-    this.hasGlobalHeader = this.$slots['header-global'] !== undefined;
+    this.hasGlobalHeader = !!(this.$slots['header-global'] !== undefined);
```
ZrianinaMariia: Also not sure if `!!` is needed here.

packages/core/src/components/cv-ui-shell/cv-side-nav-link.vue
```diff
@@ -37,7 +37,7 @@ export default {
     };
   },
   beforeUpdate() {
-    this.hasNavIcon = this.$slots['nav-icon'] !== undefined;
+    this.hasNavIcon = !!(this.$slots['nav-icon'] !== undefined);
```
ZrianinaMariia: Not sure if `!!` is needed here.

packages/core/src/components/cv-data-table/cv-data-table.vue
```diff
@@ -324,10 +324,10 @@ export default {
   methods: {
     checkSlots() {
       // NOTE: this.$slots is not reactive so needs to be managed on beforeUpdate
-      this.hasBatchActions = this.$slots['batch-actions'];
-      this.hasActions = this.$slots.actions;
-      this.hasToolbar = this.$slots.actions || this.$listeners.search || this.$slots['batch-actions'];
-      this.isHelper = this.$slots['helper-text'] !== undefined || (this.helperText && this.helperText.length);
+      this.hasBatchActions = !!this.$slots['batch-actions'];
```
ZrianinaMariia: In other places, e.g. `cv-side-nav-menu`, you use comparison with undefined `this.$slots['nav-icon'] !== undefined`. I think it makes sense to have it consistent.

packages/core/src/components/cv-data-table/cv-data-table-row.vue
lee-chase: Good point, that is a danger of find and replace. Will correct.

PR #724, opened 2019-11-29
test: add tests for cv-inline-notification

ZrianinaMariia:
#182

Tests for `CvInlineNotification`. I am having a problem with testing computed `icon` value. It should be the same as for `CvToastNotification`, but it fails with the error and I have no idea why. The test is commented.

image

#### Changelog

A       packages/core/__tests__/__snapshots__/cv-inline-notification.test.js.snap
A       packages/core/__tests__/cv-inline-notification.test.js
M       packages/core/src/components/cv-inline-notification/cv-inline-notification.vue

### DIFF
--- a/packages/core/__tests__/__snapshots__/cv-inline-notification.test.js.snap
+++ b/packages/core/__tests__/__snapshots__/cv-inline-notification.test.js.snap
@@ -1,0 +1,129 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`CvInlineNotification should render correctly 1`] = `
+<div data-notification="" role="alert" class="cv-inline-notification bx--inline-notification bx--inline-notification--error">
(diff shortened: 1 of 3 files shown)


ZrianinaMariia:
Oh, okay. `InlineNotification` import differs from `ToastNotification`, i.e.

`import WarningAltFilled20 from '@carbon/icons-vue/es/warning--filled/20';`

instead of 

`import WarningAltFilled20 from '@carbon/icons-vue/es/warning--alt--filled/20';`

@lee-chase Is it on purpose?

ZrianinaMariia:
So, first confusing came from the icon name `WarningAltFilled20` even though it was imported from `warning--filled/20`, not `warning--alt--filled/20`. Also noticed that `ToastNotification` should actually also use `warning-filled/20`, not `warning--alt--filled/20`.

PR #750, opened 2019-12-12
Fix typo in property declaration

### LINKED ISSUE #749: Typo in property of CvDataTable component
## Detailed description

The ` required`  key to the ` data`  property is misspelled. It is `requried` instead of ` required` 

Bug

`CvDataTable`

jungleBadger:
Closes #749

Fix `required` property typo in `CvDataTable` component

#### Changelog

The git diff returns nothing, this is the commit info:

```
commit 4f25b400ae2ff627084ceb68f6e6e50f3a8ed63f (HEAD -> master, origin/master, origin/HEAD)
```

### DIFF
--- a/packages/core/src/components/cv-data-table/cv-data-table.vue
+++ b/packages/core/src/components/cv-data-table/cv-data-table.vue
@@ -215,7 +215,7 @@ export default {
     sortable: Boolean,
     title: String,
     columns: { type: Array, required: true },
-    data: { type: Array, requried: true },
+    data: { type: Array, required: true },
     zebra: Boolean,
     rowsSelected: { type: Array, default: () => [] },
     helperText: { type: String, default: undefined },
--- a/packages/core/src/components/cv-data-table/cv-data-table.vue
+++ b/packages/core/src/components/cv-data-table/cv-data-table.vue
@@ -215,7 +215,7 @@ export default {
     sortable: Boolean,
     title: String,
     columns: { type: Array, required: true },
-    data: { type: Array, required: true },
+    data: { type: Array },
     zebra: Boolean,
     rowsSelected: { type: Array, default: () => [] },
     helperText: { type: String, default: undefined },


lee-chase:
Thanks, but looking at the current implementation which allows the data to be slotted I do not think this property should be required. I will adjust it along with a couple of other similar typos to be non-required.

jungleBadger:
Yeah, I do agree with your point but I thought it was intended to be `required` since it was a typo. On top of that, I think that a JSON array should also be accepted as a data source, as of today we can only pass an array with singular values. I had to use the scoped slot to achieve what I wanted.



Very often I have an array of objects on my hand rather than an array of values, so instead of pre-processing it to feed the `CvDataTable` we may come up with some mapping config between the provided JSON array and the properties that I want to draw on the table.


`["a", "b", "c"]` -> supports today

`[{"prop": "a"}, {"prop": "b"}, {"prop": "c"}]` -> sees the whole JSON as the cell value


What are your thoughts on this one? Should the lib support only the most basic form of array-values and anything different than that must be treated as slots or could we try to make this property more flexible to deal with both of the following inputs:

lee-chase:
> Yeah, I do agree with your point but I thought it was intended to be `required` since it was a typo. On top of that, I think that a JSON array should also be accepted as a data source, as of today we can only pass an array with singular values. I had to use the scoped slot to achieve what I wanted.
> 
> Very often I have an array of objects on my hand rather than an array of values, so instead of pre-processing it to feed the `CvDataTable` we may come up with some mapping config between the provided JSON array and the properties that I want to draw on the table.
> 
> What are your thoughts on this one? Should the lib support only the most basic form of array-values and anything different than that must be treated as slots or could we try to make this property more flexible to deal with both of the following inputs:
> 
> `["a", "b", "c"]` -> supports today
> 
> `[{"prop": "a"}, {"prop": "b"}, {"prop": "c"}]` -> sees the whole JSON as the cell value

I think having data that match the columns either by label or id of column (see styled columns).

First up I think we should fix the typo part, then you can raise an enhancement request.

PR #763, opened 2019-12-30
Fix property name causing reference issue

### LINKED ISSUE #762: CvDatePicker `endDate` incorrect reference
## Detailed description

At line `247` of the file ` packages/core/src/components/cv-date-picker/cv-date-picker.vue ` a type was introduced on latest changes which is causing an error at runtime

Using the `range`  type of date picker an error when selecting values:

```
v-on handler: "TypeError: Cannot read property 'endDate' of undefined"
``` 

Bug

CvDatePicker

The value reference is wrong

Anotação 2019-12-29 215158

```
"carbon-components": "^10.9.0",
"@carbon/icons-vue": "^10.8.0",
"@carbon/themes": "^10.9.0",
"@carbon/vue": "^2.20.0"
``` 

1. Import CvDatePicker component
2. Set type to range
3. Select dates

## Additional information

The issue is also happening at carbon vue storybook, so it can be reproduced here: http://vue.carbondesignsystem.com/?path=/story/components-cvdatepicker--range-using-date

jungleBadger:
Closes #762

Fix `dataValue` property reference which was causing a runtime error while trying to access `endDate` value while on range mode of the `CvDatePicker` component.

#### Changelog

M packages/core/src/components/cv-date-picker/cv-date-picker.vue

### DIFF
--- a/packages/core/src/components/cv-data-table/cv-data-table.vue
+++ b/packages/core/src/components/cv-data-table/cv-data-table.vue
@@ -215,7 +215,7 @@ export default {
     sortable: Boolean,
     title: String,
     columns: { type: Array, required: true },
-    data: { type: Array, requried: true },
+    data: { type: Array, required: true },
     zebra: Boolean,
     rowsSelected: { type: Array, default: () => [] },
     helperText: { type: String, default: undefined },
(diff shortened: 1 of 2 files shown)


lee-chase:
Thank you @jungleBadger for submitting this PR. I will add release an update next week including this fix, a canary should appear later today.

PR #868, opened 2020-04-17
fix: ensure that CvDropdown has single tabindex for accessibility

william-glasse:
Closes #868 

Ensures that the Dropdown component only has a single tab index.

#### Changelog

M packages/core/src/components/cv-dropdown/cv-dropdown.vue


lee-chase:
Thank you very much for your submission @william-glasse.

I'll add it to a canary later today and push a new latest Monday if you're happy.

PR #875, opened 2020-04-28
refactor(tag): prepare for carbon-components 10.11.2

janhassel:
Prepares the tag component for the changes made in `carbon-components@10.11.2`.
This also already includes the addition of the `high-contrast` type that is not yet in reflected in any released version of `carbon-components` but merged in `master` over in the monorepo (https://github.com/carbon-design-system/carbon/pull/5910).

#### Changelog

```
M       packages/core/src/components/cv-tag/cv-tag.vue
M       storybook/stories/cv-tag-story.js
```

Since filter tags now support to be colored as well, the `kind` prop probably needs to be deprecated and replaced by the two props `filter` (bool) and `type` (string). To avoid breaking changes I introduced a computed variable to account for uses of the current `kind` prop.
Is there any way to mark a prop deprecated so that developers would know about this change?

And one more question: do you think it's better to exclude the `high-contrast` type from this PR so that it can be merged with the 10.11.2 update or would it be better to hold back until that is release in `carbon-components`?


packages/core/src/components/cv-tag/cv-tag.vue
```diff
@@ -37,6 +39,20 @@ const componentsTags = [
   'warm-gray',
 ];
 
+const tagTypes = [
```
lee-chase, line 42: Why have we introduced this tagTypes list duplicating componentsTags which probably should be renamed tagKinds

packages/core/src/components/cv-tag/cv-tag.vue
```diff
@@ -46,15 +62,33 @@ export default {
     label: { type: String, required: true },
     kind: {
       type: String,
-      default: componentsTags[0],
       validator(val) {
         return componentsTags.includes(val);
       },
     },
+    type: {
```
lee-chase, line 69: So as to not re-use standard attributes elsewhere we have used kind for this type of switch.

packages/core/src/components/cv-tag/cv-tag.vue
```diff
@@ -46,15 +62,33 @@ export default {
     label: { type: String, required: true },
     kind: {
       type: String,
-      default: componentsTags[0],
       validator(val) {
         return componentsTags.includes(val);
```
lee-chase, line 66: Update this validator to indicate 'filter' is deprecated but permit it's use for now.

packages/core/src/components/cv-tag/cv-tag.vue
```diff
@@ -46,15 +62,33 @@ export default {
...
+      type: Boolean,
+      default: false,
+    },
   },
   computed: {
     isFilter() {
-      return this.kind === 'filter';
+      return this.filter || this.kind === 'filter';
+    },
+    tagType() {
```
lee-chase, line 85: Rename to tagKind and remove use of type

storybook/stories/cv-tag-story.js
```diff
@@ -25,6 +25,16 @@ let preKnobs = {
...
   },
+  type: {
+    group: 'attr',
+    type: select,
+    config: [
+      'Tag type',
+      ['red', 'magenta', 'purple', 'blue', 'cyan', 'teal', 'green', 'gray', 'cool-gray', 'warm-gray', 'high-contrast'],
+      'gray',
+    ],
+    prop: 'type',
```
lee-chase, line 36: Keep kind prop

Issue #758, opened 2019-12-18
Custom cv-data-table  header 


peakrock:
I want custom  cv-data-table  header ,  could you  add `slot ` for v-data-table header ?

lee-chase:
@peakrock based on the images am I right in thinking you want to be able to change the contents of the table headers cv-data-table-heading. Perhaps the label with other content?

The headings outside of columns added for expansion, batch actions and overflow menu could adopt the same pattern as data. However, sorting is currently built into each cv-data-table-heading so that might be tricky to figure out.

Feel free to submit a PR with your thoughts.

lee-chase:
@peakrock If this were possible it looks like it would be an alternative to sorting. Is this the case?

If so I think could be a candidate for core carbon @joshblack, but I'm finding it hard to justify adding complexity to the CvDataTable without it being so. 

Will leave open in case of inspiration.

lee-chase:
Been thinking about this with the submission of #890 

It may be possible to do this by 
1. Adding a slot to allow the user to define the headers
2. Promoting the header item to a full component (removing _ from name).
3. Adding slots to the header item to allow the user to put dynamic content into the label.
4. Having the header item register with the header to manage sorting where used.

Might even be nice if a tooltip could be provided without the need to add a full tooltip.

lee-chase:
@peakrock Does the canary update suit your needs?

lee-chase:
Added to previous release

PR #890, opened 2020-05-01
Add heading slot to CvDataTable component

erges:
Add an option for slotting header cells.

#### Changelog

M packages/core/src/components/cv-data-table/cv-data-table.vue
M packages/core/src/components/cv-data-table/cv-data-table-notes.md

### DIFF
--- a/packages/core/src/components/cv-data-table/cv-data-table-notes.md
+++ b/packages/core/src/components/cv-data-table/cv-data-table-notes.md
@@ -62,6 +62,26 @@ Like sorting and filtering it is the users responsibility to deal with edited da
 </cv-data-table>
 ```
 
+### Slotting header cells
+
+User can supply slotted content using the CvDataTableHeading component or plain html.
+
+```html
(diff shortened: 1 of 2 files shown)


lee-chase:
@erges it is an interesting idea to do this but I'd like to find a way that does not require person implementing the custom slot to re-implement much of CvDataTableHeading

It really depends on how much customisation of the header is required. 

One option would be to have a slot in CvDataTableHeading and pass a component down as part of the column properties. Slotting it into a fixed position.

Another would be to move much of CvDataTableHeading into a mixin and add some lifecycle function checks that verify the existence of DOM for label and sort.

The latter idea is probably more appealing along with an update to the documentation to explain what's required and a storybook example.

lee-chase:
See discussion in #758 and add your requirements there @erges 

I think I might have a plan but it would be nice to have all the requirements bottomed out in the linked issue.

lee-chase:
@erges Does the canary update suit your needs?

lee-chase:
Fixed by another PR. Thanks for the submission but further work was needed to implement this.

PR #1005, opened 2020-09-04
Remove event listener before component removal

### LINKED ISSUE #1004: [Bug] Dropdown component listener removal issue
## Detailed description

After adding a `<cv-dropdown>` component and hiding it with v-if/changing route with vue-router, there is a listener attached to `window` that keeps trying to close the dropdown, but as it doesn't exists any longer the function fails.

Bug

Dropdown

After removing the cvdropdown component, I expect the listener to be removed as well.

Version 85.0.4183.83 (Official Build) (64-bit)

"@carbon/vue": "^2.31.1",

N/A

1. Go to a page with a cv-dropdown-component rendered
2. Hide this component with v-if or using vue-router
3. Click in the screen looking the console.log error

## Additional information

This can be reproduced in the vue story book: https://vue.carbondesignsystem.com/?path=/story/components-cvdropdown--default

demoissue

jungleBadger:
Closes #1004

Call event listener removal before destroying component.

#### Changelog

{{run "git diff --name-status $(git merge-base head master)" and paste result here}}

### DIFF
--- a/packages/core/src/components/cv-dropdown/cv-dropdown.vue
+++ b/packages/core/src/components/cv-dropdown/cv-dropdown.vue
@@ -173,6 +173,9 @@ export default {
     document.body.removeEventListener('click', this.checkSlots);
     this.checkSlots();
   },
+  destroyed() {
+    document.body.removeEventListener('click', this.checkSlots);
+  },
   model: {
     prop: 'value',
     event: 'change',


jungleBadger:
Accordingly, to the [storybook](https://deploy-preview-1005--carbon-components-vue.netlify.app/?path=/story/components-cvdatepicker--default) the error still persists, I need to test it further.

packages/core/src/components/cv-dropdown/cv-dropdown.vue
```diff
@@ -173,6 +173,9 @@ export default {
     document.body.removeEventListener('click', this.checkSlots);
     this.checkSlots();
   },
+  destroyed() {
```
lee-chase, line 176: The library tends to use beforeDestroy as the Vue instance has been destroyed at this point.

Is there a good reason for it being in the destroyed hook?

lee-chase (changes requested):

packages/core/src/components/cv-dropdown/cv-dropdown.vue
jungleBadger: Thanks for the heads-up.

No special reason, I made the change and will test as soon as it is available.

lee-chase (approved):
LGTM

PR #1033, opened 2020-09-28
test: add tests for cv-tooltip

ZrianinaMariia:
#182

Add tests for CvTooltip component incl CvDefinitionTooltip and CvInteractiveTooltip. The tests for CvInteractiveTooltip do not have 100% coverage. Please feel free to suggest how to test the not-covered functionality. I spent some time on that, but it did not work at the end.

#### Changelog

A       packages/core/__tests__/__snapshots__/cv-tooltip.test.js.snap
A       packages/core/__tests__/cv-tooltip.test.js


lee-chase:
Hi Mariia, hope you are well in these rather crazy times.

Looking at the coverage report for the interactive tooltip 116/117 are only likely to be called if you are adding and removing the component, maybe through a v-if. The same is true for the beforeDestrooy function.

You could try simulating the tab key using keydown or keypress

In the same way we currently simulate clicks

  it('Raises click event when clicked', async () => {
    const wrapper = await shallow(CvIconButton);
    await trigger(wrapper.find('button'), 'click');
    expect(wrapper.emitted().click).toBeTruthy();
  });

More than happy to take your tests as an additive improvement if this is harder than I think.

ZrianinaMariia:
Hi Lee! I am good, thanks, hope the covid will be over soon. Hope you are doing well too! 

Reg. clicks - I tried several options as well as the helper `trigger` function, but for some reasons it did not work.

I also tried to test that eventListeners are attached (to check onDestroy e.g.) and I could not find a way to trigger window events, so that `position` function is called. I spied on position function and tried several options in order to trigger window event, but `position` was not called when event listeners were active.

 I will try to find some more time on the weekend to look more into that. Not sure what is better for the project and what would you prefer: merge this one or wait for the weekend. For me of course it is better to have a separate one (in case I find a way how to test the rest of the functionality) because of Hacktoberfest time (this pull request does not count actually anyway) :)

ZrianinaMariia:
I was able to add couple of more tests :)

lee-chase:
Thanks for the help @ZrianinaMariia

PR #1046, opened 2020-10-10
test: add tests for cv-toolbar

ZrianinaMariia:
#182

Add tests for CvToolbar component. Could not add coverage for @blur event for CvToolbarSearch, even though I tested that the 'active' class is removed on blur.

#### Changelog

A       packages/core/__tests__/__snapshots__/cv-toolbar.test.js.snap
A       packages/core/__tests__/cv-toolbar.test.js


lee-chase (approved):
Thanks for helping to improve our test coverage.

PR #1047, opened 2020-10-10
test: add tests for cv-code-snippet

ZrianinaMariia:
#182

Add tests for CvCodeSnippet component. Could not test toggleExpand of CvCodeSnippetMultiline partially because I was not able to import it from `@/components/cv-code-snippet`. Noticed that CvCodeSnippetMultiline props `lessText` and `moreText` are actually not reachable from CvCodeSnippet, so they can not be set.

#### Changelog

A       packages/core/__tests__/__snapshots__/cv-code-snippet.test.js.snap
A       packages/core/__tests__/cv-code-snippet.test.js


lee-chase (approved):
Great set of tests

PR #1069, opened 2020-11-03
Add after hidden modal event

### LINKED ISSUE #1068: Add "after hide" event to modal component
**Is your feature request related to a problem? Please describe.**
A common pattern I use in projects that require modals is to have a `data` property that holds an object which is displayed by the modal (either for editing or for showing information). To implement this easily I set the modals `:visible` like this:

```html
<template>
  <cv-modal :visible="objectToShow !== null" @modal-hidden="objectToShow = null"/>
</template>

<script>
export default {
  data () {
    return {
      objectToShow: null
    }
  }
}
</script>
```

This way I just have to set the object to whatever should be displayed/edited (possibly with a `v-if` on the content slot if I display data), and the object is reset when the modal is closed.

The problem is that `@modal-hidden` is triggered when the modal starts hiding, not when it is hidden. This means that there is a small amount of time when the modal is still fading out, but the content has already left. It's not a problem, but it does not look nice.

**Describe the solution you'd like**
Introducing a new event `@after-modal-hidden` (or any similar name) would solve this problem for me. It should be triggered then the CSS transition hiding the modal has finished.

**Describe alternatives you've considered**
I have two alternatives:
1) Manually attach a `transitionend` event to the correct DOM node and use that to trigger the reset (doesn't follow Vue best practices, but works)
2) Use a timeout to wait until after the modal has been hidden (which introduces magic numbers into my code)

**Additional context**
We should be able to introduce an event like this using the `transitionend` event. If I have some time left over over the next few days I'll try to implement this.

TimonLukas:
Closes #1068

Adds a new event to `cv-modal` called `after-modal-hidden`, which triggers once the hide transition is finished.

#### Changelog

M       packages/core/src/components/cv-modal/cv-modal-notes.md
M       packages/core/src/components/cv-modal/cv-modal.vue
M       storybook/stories/cv-modal-story.js

### DIFF
--- a/packages/core/src/components/cv-modal/cv-modal-notes.md
+++ b/packages/core/src/components/cv-modal/cv-modal-notes.md
@@ -66,5 +66,6 @@ This is done after the modal transitions into its visible state, which is trigge
 - modal-shown
 - modal-hidden
 - modal-hide-request - emitted when 'auto-hide-off' is set to true. Raw event with the additional attribute 'cv:reason' set to 'primary-click', 'secondary-click', 'escape-press', 'external-click' or 'close-click'
+- after-modal-hidden - emitted only after the modal hide transition is finished. It will NOT fire if the modal is hidden when it's not visible
 - primary-click
 - secondary-click
--- a/packages/core/src/components/cv-modal/cv-modal.vue
+++ b/packages/core/src/components/cv-modal/cv-modal.vue
@@ -251,5 +251,10 @@ export default {
       }
     },
   },
+  beforeDestroy() {
+    if (this.dataVisible) {
+      this.$emit('after-modal-hidden');
+    }
+  },
 };
 </script>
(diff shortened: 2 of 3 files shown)


TimonLukas:
Two things I'm not quite sure about:
1) Is this event unreliable in any environment? Ie. is there a way for the transition to not happen?
2) This event is a bit different from `modal-hidden`, which triggers after every hide request, even if the modal isn't visible. This event instead really only triggers once the hide transition has been completed. I think this is what the user would usually expect, but maybe it's just my preference.

lee-chase:
The only circumstances that the transitioned will not get triggered are if the modal is removed by conditional rendering or if the CSS gets changed.

I don't think we need to worry about either, although you could emit an event on beforeDestroy if you wanted to be sure.

TimonLukas:
Good idea! The event is only emitted if `dataVisible === true` - if I understand the component correctly, this should keep the event from being emitted when the modal is destroyed while not visible. Anything else I should look at? :)

lee-chase:
Could you add to the event to the storybook and notes file

packages/core/src/components/cv-modal/cv-modal-notes.md
/storybook/stories/cv-modal-story.js

TimonLukas:
Okay, I added it to the notes Markdown file and updated the changelog :) I already added the event to the story as an action in the initial PR, is there something left to add?

lee-chase (approved):
Thanks for the PR.

PR #1081, opened 2020-11-25
fix: ensure no previous transitionend event triggers after-modal-hidden

### LINKED ISSUE #1080: cv-modal after-modal-hidden event sometimes triggers too early
## Detailed description

I've been using the `after-modal-hidden` event I implemented in #1069 for a bit, and when using it in a real project I noticed that the event is sometimes emitted too early.

It's a bug - the event should always be fired after the modal has been hidden, never earlier.

`cv-modal`

Chrome

2.35.0

Nothing pressing :)

I'm not sure about a reliable reproduction, I can't get the same bug to appear in the storybooks.

## Additional information

If needed I can provide an IBM-internal Github repository in which this bug occurs.

TimonLukas:
Closes #1080 

Currently a `border-bottom` transition can trigger the event early. This fix ensures that the event is only triggered after the modal is no longer visible.

#### Changelog

M       packages/core/src/components/cv-modal/cv-modal.vue

### DIFF
--- a/packages/core/src/components/cv-modal/cv-modal.vue
+++ b/packages/core/src/components/cv-modal/cv-modal.vue
@@ -245,14 +245,17 @@ export default {
       document.body.classList.remove(`${this.carbonPrefix}--body--with-modal-open`);
 
       if (this.dataVisible) {
-        this.$el.addEventListener('transitionend', this.afterHide, { once: true });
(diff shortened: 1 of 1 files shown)


lee-chase (approved):
LGTM thank you for the submission Timon.

PR #1082, opened 2020-11-25
Include `left` and `right` to direction validator.

jungleBadger:
As of today, the definition tooltip only accepts `top` and `bottom`. To remain consistent with the `cv-tooltip` definition, I included `left` and `right` options.

Plus, left and right behavior is applied, but the validator function throws an error to the console.


lee-chase (approved):
Thank you for the submission @jungleBadger  looks good to me

PR #1091, opened 2020-12-07
fix: change update to computed

davidnixon:
Closes #

The side nav with icon does not work properly if the side nav is hidden with v-if and then later shown
https://ihzer.csb.app/
image

Force an update to show icons
image

#### Changelog

M       packages/core/src/components/cv-ui-shell/cv-side-nav-link.vue

### DIFF
--- a/packages/core/src/components/cv-ui-shell/cv-side-nav-link.vue
+++ b/packages/core/src/components/cv-ui-shell/cv-side-nav-link.vue
@@ -34,13 +34,10 @@ export default {
     active: Boolean,
     icon: Object,
   },
-  data() {
-    return {
-      hasNavIcon: this.$slots['nav-icon'],
-    };
-  },
-  updated() {
-    this.hasNavIcon = !!this.$slots['nav-icon'];
+  computed: {
+    hasNavIcon() {
+      return !!this.$slots['nav-icon'];
+    },
   },
 };
 </script>


packages/core/src/components/cv-ui-shell/cv-side-nav-link.vue
```diff
@@ -34,13 +34,10 @@ export default {
...
   },
-  data() {
-    return {
-      hasNavIcon: this.$slots['nav-icon'],
-    };
-  },
-  updated() {
-    this.hasNavIcon = !!this.$slots['nav-icon'];
+  computed: {
+    hasNavIcon() {
```
lee-chase, line 38: Slots are not reactive so I am not sure this will work. Generally, I tend to use the following pattern. Does this work for you?

```
data(): {
  hasSlotX: false
},
mounted() {
  this.checkSlots();
},
updated() {
  this.checkSlots();
},
methods: {
  checkSlots() {
    this.hasSlotX = !!this.$slots['slot-x'];
  },
}
```

davidnixon: Yeah the slots not being reactive is the core issue.

Both the computed version and your version with the updated and mounted both work. 

Testing with my use case,  I added some logging and I can see that adding the check to the mounted function fixes the issue. The update function is not called. 

I can also see that if I use the computed version, the hasNavIcon() function is called during the mount process and update process.

LMK if you want me to change this to use the updated/mounted pattern above. I didn't look at the other file cv-side-nav* files but now that I am looking I see they are all using that updated/mounted pattern so probably better that they are all the same.

lee-chase:
Closing in favour of #1902

PR #1094, opened 2020-12-16
feat: add show/hide events to cv-interactive-tooltip

davidnixon:
We had a use case in our app to animate a part of the page when the tooltip is shown and when the tooltip is hidden. There is not an easy way to do this without events emitted from the tooltip so I forked the component to use it in our app.

ToolTipEvent

#### Changelog

M       packages/core/src/components/cv-tooltip/cv-interactive-tooltip.vue
M       packages/core/src/components/cv-tooltip/cv-tooltip-notes.md

### DIFF
--- a/packages/core/src/components/cv-tooltip/cv-interactive-tooltip.vue
+++ b/packages/core/src/components/cv-tooltip/cv-interactive-tooltip.vue
@@ -105,6 +105,13 @@ export default {
         this.position();
       }
     },
+    dataVisible() {
+      if (this.dataVisible) {
+        this.$emit('show-tooltip');
+      } else {
+        this.$emit('hide-tooltip');
+      }
+    },
   },
   methods: {
     positionListen(on) {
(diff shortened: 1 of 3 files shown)


lee-chase (changes requested):
Looks good, could you add actions in storybook?

davidnixon:
> Looks good, could you add actions in storybook?

done

davidnixon:
> > Looks good, could you add actions in storybook?
> 
> done

Actually let me re-work this to just have a show event with a true/false param. That would be cleaner.

davidnixon:
> > > Looks good, could you add actions in storybook?
> > 
> > 
> > done
> 
> Actually let me re-work this to just have a show event with a true/false param. That would be cleaner.

😃  OK last change. Really this time. I looked for other examples and found CVModal which also has a show/hide event so I used that as a model instead. It's still 2 different events but slightly renamed to match the CvModal pattern.
https://github.com/carbon-design-system/carbon-components-vue/blob/d5f18bdc0a76ff9e792765f799faba5b164693fb/packages/core/src/components/cv-modal/cv-modal.vue#L210
https://github.com/carbon-design-system/carbon-components-vue/blob/d5f18bdc0a76ff9e792765f799faba5b164693fb/packages/core/src/components/cv-modal/cv-modal.vue#L252

Issue #1095, opened 2020-12-16
cv-multi-select opens with scroll on second element


gabriel-lavaqui:
## cv-multi-select opens with scroll on second element

When expanded, cv-multi-select opens with scroll on the second element, hiding the first.

### Describe in detail the issue you're having.
The first element of the list in cv-multi-select is hidden by the scroll when it opens.
It happens on most tries.

### Is this a feature request (new component, new icon), a bug, or a general issue?
This is a bug.

### Is this issue related to a specific component?
Yes, `cv-multi-select`.

### What browser are you working in?
Chrome version 87.0.4280.88 on Mac OS

### What version of the Carbon Design System are you using?

```
"@carbon/vue": "^2.35.0"
"carbon-components": "^10.25.0"
```

## Steps to reproduce the issue

1. Go to the [story of cv-multi-select](http://vue.carbondesignsystem.com/?path=/story/components-cvmultiselect--default)
2. Click to expand the list
3. Notice that several times you can scroll up to reveal the first element

## Additional information

![multi-select-bug](https://user-images.githubusercontent.com/25914871/102390200-dc702e00-3fb2-11eb-8c79-8cc87d31e4ea.gif)

PR #1098, opened 2020-12-18
Fix scroll behavior

gabriel-lavaqui:
Closes [#1095](https://github.com/carbon-design-system/carbon-components-vue/issues/1095)

Fixes the behavior where each other click to open the cv-multi-select component hide the first element

#### Changelog

M       packages/core/src/components/cv-multi-select/cv-multi-select.vue


lee-chase:
Thank you very much for this improvement.

PR #1104, opened 2021-01-14
fix: cv-date-picker range plugin  use

lee-chase:
When swapping out the range plugin for the carbon version left a spurious 'new' which did not show one storybook. Can be seen here with v2.36.0 https://codesandbox.io/s/carbonvue-cvdatepicker-i1lyy?file=/src/App.vue

#### Changelog

m. packages/core/src/components/cv-date-picker/cv-date-picker.vue


Fontinalis (approved):
Nice! So that caused my issue... I had this issue, but due to some strict timeline, my "quick fix" was to use two single cv-date-picker component. Thank you for the fix!

PR #1119, opened 2021-02-13
v3: experimenting with CvTag component, including stories

Fontinalis:
Hey @lee-chase,

Seen your commits and decided to join you in the effort of moving forward with the support for Vue 3.

In this PR, I've been experimenting on adding the CvTag component to the v3 package, including stories for documentation. Just something small to begin with.

I see a huge potential in refactoring these components and making them better on the way, extending their features with ones that are already being used in the original React package, but missing from the current `@carbon/vue` package.

Let me know if you need me to change something, I'm happy to do so. 

#### Changelog

```
M       packages/v3/package.json
A       packages/v3/src/components/CvTag/CvTag.stories.js
A       packages/v3/src/components/CvTag/CvTag.vue
A       packages/v3/src/components/CvTag/__tests__/CvTag.spec.js
A       packages/v3/src/components/CvTag/consts.js
A       packages/v3/src/components/CvTag/index.js
```

PR #1135, opened 2021-02-28
Fontinalis v3/cv tag

lee-chase:
Closes #1119

Includes PR from Fontinails of CvTag with a few tweaks to match the latest Storybook beta format.

Also catches missed code line during test.


Fontinalis:
Thank you for finishing it! Sorry for the late response, got busy with work and other things. You seem like having a great progress on moving forward with vue3, will try to contribute more as soon as I'll have a bit more time.

PR #1145, opened 2021-03-18
fix: change label HTML tag in cv-file-uploader component

caioafc:
Closes #

Hello guys,
We have an accessbility issue in cv-file-uploader component label, `<strong/>` tag is not the best option for heading contents. To fix this issue I tried to change to `<p/>` tag, so the Accessbility Checker stoped warning about this problem. Can you check if the alteration is possible?

### Before: 
image

### After:
image

#### Changelog

`M       packages/core/src/components/cv-file-uploader/cv-file-uploader.vue`


jungleBadger (commented):
It does look good to me. @lee-chase, was there a specific reason to use the `strong` tag?

lee-chase:
This change would need to go through core carbon as our HTML mirrors that.

Raised an issue with core Carbon.

lee-chase (approved):
Looks good. Carbon have gone with this solution.

PR #1150, opened 2021-04-01
v3 CvBreadcrumb

TimonLukas:
Implements CvBreadcrumb & CvBreadcrumbSkeleton incl. stories and tests. Same HTML as in current React release while a bit more flexible.

Only problem I currently see is that the skeleton component in the docs doesn't show the right code ([Storybook Issue](https://github.com/storybookjs/storybook/issues/14110)). We might have to separate it into its' own story, since it's a functionally completely separate component maybe not a bad idea.


lee-chase (approved):
LGTM

PR #1152, opened 2021-04-04
v3 CvSkeletonText

TimonLukas:
Implements CvSkeletonText incl. stories and tests. Same HTML as in current React release.

The way the width of the individual lines is computed is changed from the old implementation - the React implementation always produces a repeating pattern of 3 values, which I've replicated. I'm not sure it's intended by their implementation though, do you know that @lee-chase?

Additionally I've dropped the `paragraph` prop - it's a bit confusing why there's both a `paragraph` and a `lineCount`:
- If `paragraph === false`, `lineCount` doesn't do anything
- If `lineCount === 1`, `paragraph = true` only makes the element slightly less wide

Now we only have `lineCount`. If `lineCount === 1` we replicate the behavior of `paragraph === false`, otherwise we replicate `paragraph === true`.


packages/v3/src/components/CvSkeletonText/CvSkeletonText.vue
```diff
@@ -0,0 +1,67 @@
...
+      :style="{ width: widths[i - 1] }"
+    />
+  </div>
+</template>
+
+<script>
+import { computed } from 'vue';
+import { carbonPrefix } from '../../global/settings';
+
+const LENGTH_MODIFIERS = [`64px`, `2px`, `32px`];
```
lee-chase, line 27: Is there a good reason to use these values instead of the Carbon react method?

```
const randoms = [0.973051493507435, 0.15334737213558558, 0.5671034553053769];

function getRandomInt(min, max, n) {
  return Math.floor(randoms[n % 3] * (max - min + 1)) + min;
}
```

lee-chase:
The pattern of 3 looks intended, and good to get rid of the `paragraph` property. Just one question about the line length calculations.

lee-chase:
Tests missed line 52 @TimonLukas ` CvSkeletonText.vue           |    90.91 |       50 |      100 |    90.91 |                52`

packages/v3/src/components/CvSkeletonText/CvSkeletonText.vue
TimonLukas: I chose to implement it more naively because the Carbon React method kind of confused me at first. My implementation replicates the behavior while being more straight-forward, but there's no reason not to go with the React implementation :)

lee-chase: Fair enough

TimonLukas:
Good catch @lee-chase! I had that guard left in from earlier and forgot to remove it once it was dead.

lee-chase (approved):

PR #1153, opened 2021-04-04
v3 CvAspectRatio

TimonLukas:
Implements CvAspectRatio incl. stories and tests. Same HTML as in current React release.

I added the React versions `as` prop.


packages/v3/src/components/CvAspectRatio/CvAspectRatio.vue
```diff
@@ -0,0 +1,31 @@
...
+export default {
+  props: {
+    as: {
+      type: String,
+      default: 'div',
+    },
+    ratio: {
+      type: String,
+      default: aspectRatios[0],
+      validator: value => aspectRatios.includes(value),
```
lee-chase, line 26: Previously there was an informative console.warn when an invalid aspect ratio was supplied. Button I notice does console.error. 

Unless. you are suggesting no warning should be given then I think we should be consistent as error or warning.

NOTE: Mock console (as per button) if outputting a message.

TimonLukas: I think a more descriptive error message would be great, Vue's are never that good for custom props. I like the implementation in `CvButtonCommon`, although we should probably always return `true` - otherwise two errors are printed per property, and if we implement our own more explicit one we should hide theirs.

It would probably then also make sense to make this a reusable pattern, I'll take a look at that this week.

PR #1154, opened 2021-04-04
v3 CvInlineNotification & CvToastNotification

TimonLukas:
Implements CvInlineNotification & CvToastNotification incl. stories and tests. Same HTML as in current React release.

### DIFF
--- a/packages/v3/src/components/CvNotification/CvToastNotification.vue
+++ b/packages/v3/src/components/CvNotification/CvToastNotification.vue
@@ -1,0 +1,84 @@
+<template>
+  <div
+    role="alert"
+    :kind="kind"
+    :class="[
+      `${carbonPrefix}--toast-notification`,
+      `${carbonPrefix}--toast-notification--${kind}`,
+      { [`${carbonPrefix}--toast-notification--low-contrast`]: lowContrast },
+    ]"
+  >
+    <component
+      :is="icon"
+      :class="`${carbonPrefix}--toast-notification__icon`"
+    />
+    <div :class="`${carbonPrefix}--toast-notification__details`">
+      <h3 :class="`${carbonPrefix}--toast-notification__title`">{{ title }}</h3>
+      <div :class="`${carbonPrefix}--toast-notification__subtitle`">
+        {{ subTitle }}
+      </div>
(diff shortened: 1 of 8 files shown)


packages/v3/src/components/CvNotification/CvToastNotification.vue
```diff
@@ -0,0 +1,84 @@
...
+      type: String,
+      default: 'Close notification',
+    },
+    /** Use lower contrast version of the notification */
+    lowContrast: {
+      type: Boolean,
+      default: false,
+    },
+  },
+  emits: ['action', 'close'],
```
lee-chase, line 77:
```suggestion
  emits: [
    /**
     * action: emitted on clicking the action button
     */
    'action',
    /**
     * close: emitted one close
     */
    'close',
  ],
```

packages/v3/src/components/CvNotification/CvToastNotification.vue
```diff
@@ -0,0 +1,84 @@
...
+    />
+    <div :class="`${carbonPrefix}--toast-notification__details`">
+      <h3 :class="`${carbonPrefix}--toast-notification__title`">{{ title }}</h3>
+      <div :class="`${carbonPrefix}--toast-notification__subtitle`">
+        {{ subTitle }}
+      </div>
+      <div :class="`${carbonPrefix}--toast-notification__caption`">
+        {{ caption }}
+      </div>
+    </div>
```
lee-chase, line 23: The react version seems to allow children after subtitle and caption. We should add a slot.

packages/v3/src/components/CvNotification/CvToastNotification.vue
TimonLukas: Is there a reason to put the JSDocs in 3 lines instead of 1? Otherwise I'd suggest to go with this:

```javascript
emits: [
  /** Emitted on clicking the action button */
  'action',
  /** Emitted on close */
  'close',
]
```

What do you think?

lee-chase: No reason to go with 3 lines, it's just a VSCode habit I've fallen into. 1 Line looks much neater.

lee-chase (approved):

Issue #1161, opened 2021-04-07
Sortable prop issue on cv-data-table 


techtolentino:
## Detailed description

Setting `sortable` to `false` is revealing the sortable table headers

> What browser are you working in?
Chrome and Firefox (latest)

> What version of the Carbon Design System are you using?

`"@carbon/vue": "^2.37.1"`
`"carbon-components": "^10.32.0"`
`"nuxt": "^2.14.12"`

Screen capture of what I'm seeing:

https://user-images.githubusercontent.com/6276074/113893927-bac6cc80-9784-11eb-8588-daabf71680ee.mov

TimonLukas:
Hey! Thank you for the report. You're setting sortable to the string `"false"`, which is a truthy value. Would you mind trying `:sortable="false"`? 🙂

techtolentino:
@TimonLukas - okay, this makes sense and it works (🤦 )

Thank you for your quick response, too, btw.

The reason why I was digging into the `sortable` prop, is because for some reason, the `[aria-sort]` attr selector is being used to set table header paddings to `padding: 0`

But when `:sortable="true"`, the padding is there... 

Maybe this is the real issue to file, or is there something else I'm missing?

TimonLukas:
Glad to hear that it works now!

`[aria-sort]` is used as the selector because if `sortable === true` the table headings are placed into buttons, which have their own styles for centering the text, or at least I think so based on a quick look.

To me this behavior seems correct, I'm afraid I didn't understand what issue you're referring to at the end. Could you describe the issue using the [CvDataTable story](https://vue.carbondesignsystem.com/?path=/story/components-cvdatatable--default)? Thank you!

techtolentino:
Apologies - I didn't mean to close it.

I'll show you what I'm seeing in the CvDataTable story:

When i set `sortable` to `false` (in knobs), you'll see the TableHeader losing its padding:


https://user-images.githubusercontent.com/6276074/114016918-9f150200-9828-11eb-81bd-7272aea84fdf.mov



is this expected behavior?

TimonLukas:
Ah, sorry, I get it now! No, I don't think this is supposed to happen, at least the React implementation [doesn't show this behavior](https://react.carbondesignsystem.com/?path=/story/components-datatable--basic-table).

[...]

PR #1162, opened 2021-04-08
fix: cv-data-table headers have no padding when sortable = false

TimonLukas:
Contributes to #1161

## What did you do?
I fixed `CvDataTable` headers having no padding when they are not sortable, but the table has the sortable class (e.g. when only some headers are sortable).

## Why did you do it?
Fix a bug

## How have you tested it?
I reproduced the error as described in the issue using the Storybook

- [x] N/A
- [ ] No
- [ ] Yes


lee-chase (approved):

mesenev:
Where can we expect release with the fix?

PR #1167, opened 2021-04-09
[v3] chore(deps): add @vueuse/core dependency

TimonLukas:
## What did you do?
Added `@vueuse/core` as a dependency for v3

## Why did you do it?
It seems like a great collection of re-usable Vue tools, which will lower our own implementation complexity.

### DIFF
--- a/packages/v3/package.json
+++ b/packages/v3/package.json
@@ -21,6 +21,7 @@
     "@carbon/layout": "^10.17.0",
     "@carbon/themes": "^10.26.0",
     "@carbon/type": "^10.20.0",
+    "@vueuse/core": "^4.8.1",
     "carbon-components": "10.30.0",
     "core-js": "^3.6.5",
     "flatpickr": "^4.6.9",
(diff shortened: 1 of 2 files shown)


lee-chase:
Which parts of @vueuse/core were you thinking of using?

TimonLukas:
Many functions could prove useful for us:
- [useDebounceFn](https://vueuse.org/shared/useDebounceFn/) for reactive debouncing (see my implementation of [`_CviCopyButton`](https://github.com/TimonLukas/carbon-components-vue/blob/71762d665d103b6e3209ac0ce55d1c54ac010964/packages/v3/src/components/CvCopyButton/_CviCopyButton.vue))
- [onClickOutside](https://vueuse.org/core/onClickOutside/) could be used for modals
- [useElementBounding](https://vueuse.org/core/useElementBounding/) will be very useful for code snippets
- [useClipboard](https://vueuse.org/core/useClipboard/) for code snippets
- [useVModels](https://vueuse.org/core/useVModels/) for components with v-models

This is just from a quick overview. I think there are a lot of functions in there that will make our lives a lot easier, and it doesn't have any sub-dependencies.

hedphuqz:
Hi, so I also had a look at vue-use and I much prefer the idea of setting up v-models this way. It tightly couples our v-model names with our prop names which helps bring more uniformity to the code we're writing. 

I think also that utilities such as `useElementBounding` could simplify quite a lot of our work, especially when looking at tooltips for example.

I think one of the bigger advantages is it makes our code much more declarative and therefore easier to read and maintain.

lee-chase (approved):
OK. Lets use @vueuse/core

PR #1175, opened 2021-04-18
v3 CvCodeSnippet & CvCopyButton

TimonLukas:
Adds v3 CvCodeSnippet & CvCopyButton. Full test coverage, only the `use` hooks are missing a bit. I'll improve upon this later when I better know how they'll be used, let me know if it blocks merging for you.


packages/v3/src/use/components.js
```diff
@@ -0,0 +1,27 @@
...
+    reactivePick(props, Object.keys(subcomponents[props[kindKey]].props))
+  );
+
+/**
+ * Reactive subcomponent and sub props
+ * @param props Reactive super component props
+ * @param subcomponents Map of kind name to component
+ * @param kindKey key under which to find the current kind name in props
+ */
+export const useSubcomponent = (props, subcomponents, kindKey = 'kind') => ({
```
lee-chase, line 21: I am a bit conflicted here. In the Vue2 version, we unused props would be passed through as $attrs and by default bound to the component. Here I believe we are discarding the props.

lee-chase (changes requested):
@TimonLukas Looks good. One change request and one item for discussion.

packages/v3/src/components/CvCodeSnippet/_CviCodeSnippetOneline.vue
```diff
@@ -0,0 +1,107 @@
...
+      {
+        [`${carbonPrefix}--snippet--no-copy`]: hideCopyButton,
+        [`${carbonPrefix}--snippet--disabled`]: disabled,
+        [`${carbonPrefix}--snippet--light`]: light,
+      },
+    ]"
+  >
+    <div
+      role="textbox"
+      :tabindex="disabled ? undefined : 0"
```
lee-chase, line 15: Annoyingly this offends my editor - maybe there is an update.

image

Alternatively this (which is not as clear) does not `v-bind="disabled ? {} : { tabindex: 0 }"`

TimonLukas: Based on the MDN article we could probably set it to `-1` in the disabled case. This means that it's not reachable through tabbing, which seems to be the intention.

packages/v3/src/use/components.js
TimonLukas: We are indeed discarding the additional props, since adding them would lead to rendered HTML like this:

```html
<button type="button" class="bx--copy bx--snippet bx--snippet--inline" aria-live="polite" aria-label="" aria-describedby="8b2dc7ff-0934-498f-b0f4-c92679fbf0b2" kind="inline" wraptext="false" lesstext="Show less" moretext="Show more" maxcollapsednumberofrows="15" maxexpandednumberofrows="0" mincollapsednumberofrows="3" minexpandednumberofrows="16"><code id="8b2dc7ff-0934-498f-b0f4-c92679fbf0b2">node -v</code> Copied! <span aria-hidden="true" class="bx--assistive-text bx--copy-btn__feedback">Copied!</span></button>
```

Any attributes that are not part of the parent component props will automatically be applied to the resulting outer element. E.g. if you do `<cv-code-snippet kind="inline" asdf>`, the rendered `<button type="button"...>` would have the `asdf` attribute.

The only thing that can't be passed are attributes that are defined on the parent, but not on the child - but I think it's okay if we lose those.

lee-chase: Understood.

PR #1179, opened 2021-05-03
Bugfix: Add missing Classes to CvTextInput (Password Visibility Toggle)

floriankapaun:
Contributes to `CvTextInput`

## What did you do?

I conditionally added classes that are apparently needed for the password visibility toggle

## Why did you do it?

When working with VueCarbonComponents I noticed that the password visibility toggle was missing some styles. When comparing the Vue Component with its corresponding React Component I noticed that those classes were missing although the styles were already included.

Bildschirmfoto 2021-05-03 um 16 36 23

## How have you tested it?

I ran the command included in package.json (`yarn test`). I don't know if this is sufficient, but I couldn't find any information on the correct procedure. Updated the Snapshot as well.

- [X] No, no update needed


lee-chase (approved):

lee-chase:
Looks great thanks for the PR.

PR #1231, opened 2021-07-28
fix(file-uploader): mark files with invalid types/formats

### LINKED ISSUE #1223: File Uploader Error Scenario
## Detailed description

> When we upload a wrong file, Error message should be displayed. But it is not showing as expected

This is a bug

CvFileUploader

Error message should be display as design

Chrome

10.23.1

1. Open link : https://vue.carbondesignsystem.com/?path=/story/components-cvfileuploader--default
2. Try uploading wrong file 
3. Error message is not showing

## Additional information

- Screenshots or code
- 
image

- Notes
Expected behavior should be 
image

trickstival:
Fix #1223 

## What did you do?

- made the accept attr a prop
- checked if dropped files have the same mime type or extensions as the informed in the accept prop
- added special case for jpeg files

## Why did you do it?
because there was a comment there saying it should be implemented and an issue for it.
I inspired my code in the [carbon-react implementation](https://github.com/carbon-design-system/carbon/blob/0dfde60e30a3e1238d49b66950362fd39a2c7ca7/packages/react/src/components/FileUploader/FileUploaderDropContainer.js#L65)

## How have you tested it?
I created unit tests and tested with the local storybook

- [ ] N/A
- [x] No
- [ ] Yes

### DIFF
--- a/packages/core/__tests__/cv-file-uploader.test.js
+++ b/packages/core/__tests__/cv-file-uploader.test.js
@@ -1,0 +1,87 @@
+import { testComponent, awaitNextTick } from './_helpers';
+import { CvFileUploader } from '@/components/cv-file-uploader';
+
+const { shallowMount: shallow, trigger, setProps } = awaitNextTick;
+
+describe('CvFileUploader', () => {
+  // ***************
+  // FUNCTIONAL TESTS
+  // ***************
+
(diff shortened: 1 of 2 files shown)


lee-chase (approved):
Thanks for your PR

PR #1247, opened 2021-08-25
Initial search value in data table

MichaelAmorim4:
Contributes to [carbon-components-vue](https://github.com/carbon-design-system/carbon-components-vue)

## What did you do?

Create a prop to allow the initial search value of the search bar in the data table to be set

## Why did you do it?

So the table can be filtered at the start and the user can see what the filter is in the search bar

- [x] Yes


lee-chase (approved):

lee-chase:
@MichaelAmorim4 approved although it would be good to add the new prop to storybook before merging.

PR #1258, opened 2021-10-01
fix: Do not display empty tooltip in `cv-icon-button`

mateusbandeiraa:
## What did you do?
1. Do not display a tooltip in `cv-icon-button` if `label` is empty
2. Add a temporary fix to display `cv-icon-button`s correctly inside `cv-header-global-action`

## Why did you do it?
`cv-icon-button` has a tooltip that displays the label value:
image
Currently, if `label` is empty, an empty tooltip will be displayed:
image
This can be specially problematic in `cv-ui-shell`, where we might not want to add a tooltip to the header action buttons:
image

### About the 'temporary fix'
The CSS rule that aligns the SVG icon in a `cv-icon-button` inside a `cv-header-global-action` is defined as
```scss
 .#{$prefix}--header__action.#{$prefix}--btn--icon-only.#{$prefix}--tooltip__trigger { 
   justify-content: center; 
 } 
```
(https://github.com/carbon-design-system/carbon/blob/354bf5c43d816485d602bef68e336911105f3829/packages/styles/scss/components/ui-shell/header/_header.scss#L82-L84)

This implies that only `cv-icon-buttons` with a tooltip class will be selected. Hence, buttons without a tooltip would be misaligned. I opened an issue on the main carbon repository (carbon-design-system/carbon#9789), but meanwhile, I made a temporary fix on our side.

```Vue
<button
    ...
    :class="[...buttonClassOpts({ iconOnly: true }), ...iconButtonClassOpts({ hasTooltip: !!label })]"
    ...
  >
```
```JavaScript
methods: {
    iconButtonClassOpts(opts = {}) {
      const classes = [];
      if (opts.hasTooltip) {
        // ...
      } else {
        classes.push(`tooltip-temporary-fix`);
      }
      return classes;
    },
  },
```
```HTML
<style>
/* see carbon-design-system/carbon#9789 */
.tooltip-temporary-fix {
  display: flex;
  justify-content: center;
}
</style>
```

I'm not sure if this is a good solution-- maybe we should wait until the issue I opened in carbon-design-system/carbon#9789 is discussed further.

## How have you tested it?
I ran `yarn test` and updated the relevant tests.

- [x] N/A
- [ ] No
- [ ] Yes


lee-chase:
The React implementation raises an error if no iconDescription is supplied. Any update here should really add a validator to try and enforce usage of the icon button.

image

mateusbandeiraa:
The react implementation doesn't use `renderIcon` on the button rendered by `HeaderGlobalAction` (https://github.com/carbon-design-system/carbon-components-react/blob/c3bf0123a2a98ca84acfc7e86e69840eaf57a89e/src/components/UIShell/HeaderGlobalAction.js). It instead passes the icon as a child element to be placed inside the `<button>` element. That's why there's no error thrown here.

As you can see in [https://codesandbox.io/s/header-with-actions-vh014](https://codesandbox.io/s/header-with-actions-vh014), header actions won't show tooltips.

PR #1271, opened 2021-10-29
fix: error due to l10n exported with "default" field.

OhotnikovM:
Hello!
We have an error due to this part of code because it is outside the component and after build it causes error because this code use global l10n module in bundle with nested `default` filed. That is we should apply it like `l10n.default.en` and etc or replace this code into component module.
Our team would be eternally grateful if you approve this pull request!

Contributes to #
- carbon-components-vue

## What did you do?
- We have an error due to this part of code because it is outside the component and after build it causes error because this code use global l10n module in bundle with nested `default` filed. That is we should apply it like `l10n.default.en` and etc or replace this code into component module.

## Why did you do it?
- replaced part of code inside component module from global scope

## How have you tested it?
- i did it locally and tested

- No
- [ ] N/A
- [x] No
- [ ] Yes


lee-chase (approved):

OhotnikovM:
@lee-chase Hello! It seems that i can't merge it myself due to protected branch. Could you merge it please?

lee-chase:
Will do. Just waiting for the ci/build to finish. Don't know why ci/circleci is still there.

OhotnikovM:
I got it. Thank you so much!

lee-chase (approved):

PR #1286, opened 2021-12-21
fix: change the aria role of the list element

### LINKED ISSUE #1255: CvTabs: ARIA attribute 'aria-selected' is not valid for the element <li> with ARIA role 'none'
Access Accessibility Checker found an issue in the CvTabs component.

The text says **ARIA attribute 'aria-selected' is not valid for the element 'li' with ARIA role 'none'** but according to the screenshot the role is "presentation. I don't know why.

I changed the role in the debugger to 'tab' to see whether the issue could be fixed this way and then the issue wasn't reported by the Checker anymore

image

rodet:
Fixes #1255

## What did you do?
Change the role of the `<li>` element from `presentation` to `tab`.

## Why did you do it?
This would be consistent with the Carbon React components and fix the accessibility issue documented by #1255.

## How have you tested it?
Not yet well enough. Thus creating this as a draft PR so it doesn't get forgotten over the holidays.

- [X] N/A
- [ ] No
- [ ] Yes

### DIFF
--- a/packages/core/src/components/cv-tabs/cv-tabs.vue
+++ b/packages/core/src/components/cv-tabs/cv-tabs.vue
@@ -43,7 +43,7 @@
               [`${carbonPrefix}--tabs--scrollable__nav-item--selected`]: selectedId == tab.uid,
             },
           ]"
-          role="presentation"
+          role="tab"
           :aria-selected="selectedId == tab.uid ? 'true' : 'false'"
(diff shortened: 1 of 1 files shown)


rodet:
@lee-chase what's in the way of merging this? Any way I can help maybe?

lee-chase:
> @lee-chase what's in the way of merging this? Any way I can help maybe?

Not sure if I'm missing something @rodet #1319 makes a similar change but v10 carbon still looks like this.

image

PR #1302, opened 2022-03-28
feat: [V3] add CvTextInput

mateusbandeiraa:
I'm implementing `CvTextInput` on V3.

I'm not done yet, but feedback is welcome. Actually it would be great because this would be my first contribution to V3.

### DIFF
--- a/packages/v3/src/components/CvTextInput/CvTextInput.stories.js
+++ b/packages/v3/src/components/CvTextInput/CvTextInput.stories.js
@@ -1,7 +1,6 @@
-import CvTextInput from './CvTextInput';
+import CvTextInput from './CvTextInput.vue';
 import {
   sbCompPrefix,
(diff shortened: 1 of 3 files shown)


mateusbandeiraa:
This PR depends on #1303 because of the newer `eslint-plugin-vue` version. I've rebased this branch on top of 1303, but will rebase back to the main branch once the plugin update is merged.

mateusbandeiraa:
I think we'll need to update some more dependencies. 

The `<script setup>` syntax was releases as stable on Vue 3.2; I think we should raise the dependency version to `^3.2.0`.

Also, the current Storybook version wouldn't work when I used `defineEmits()` with the new syntax. Running `npx sb@latest upgrade` fixed the issue.

There are some other dependencies that I'm not quite sure are necessary. I'll open an issue to discuss them later.

PR #1319, opened 2022-04-22
fix: li and button role issues

travis1111:
The roles on LI and Button are wrong.

UL has role=tablist, LI should have role=tab and button should have role=button. 

- [ ] N/A
- [ x] No
- [ ] Yes


lee-chase:
You may be correct @travis1111 but the Carbon 10 implementation has the following format

image

PR #1329, opened 2022-05-08
chore: switch acc  to script setup

lee-chase:
Slack question regarding use of script setup

## What did you do?

Changed the vNext accordion component to script setup

## Why did you do it?

As an experiment for comparison.

## How have you tested it?

Only in storybook as the tests seem to be down

image

- [x] N/A
- [x] No
- [ ] Yes


davidnixon:
It looks like maybe I missed to check something in. I'll have a look and get this merged today.

src/components/CvAccordion/CvAccordion.vue
```diff
@@ -14,79 +14,80 @@
...
-    provide('onAccItemChagne', (clickedItemCvId, open) => {
-      items.get(clickedItemCvId).open = open;
-      emit('change', {
-        change: { id: clickedItemCvId, open },
-        state: state.value,
-      });
-    });
+const emit = defineEmits({
+  /**
+   * Triggers when the clear button is pressed on a filter CvTag
```
davidnixon, line 48: Comment is off

davidnixon:
For me the storybook works. I cloned to a new directory and ran `yarn` and then `yarn serve`
I'm on Linux so maybe that's the difference? I'll try on osx tomorrow.

The accordion story book looks like this:
image

It all looks good except for the description of the change event 
> Triggers when the clear button is pressed on a filter **CvTag**

Also the story shows slots but there are no slots for the accordion. I do not know storybook well enough yet to know if that is expected or not.

davidnixon:
I'll look to adding the test scripts to the actions for the branch.

Issue #1343, opened 2022-07-18
 Accessibility in cv-dropdown-item


barbarabeat:
## Detailed description

> Describe in detail the issue you're having.

Run IBM Equal Access Accessibility Checker on the cv-dropdown-item, result Violation: “”The element with role "menuitemradio" is not contained in or owned by an element with one of the following roles: "group, menu, menubar””

Cv-dropdown-item don not inserts a role in the ‘li’ element.

> Is this a feature request (new component, new icon), a bug, or a general issue?

accessibility issue

> Is this issue related to a specific component?

cv-dropdown-item

> What did you expect to happen? What happened instead? What would you like to see changed?

Cv-dropdown-item creates an role=“menuitem” in the “li”. 

> What browser are you working in?

firefox

> What version of the Carbon Design System are you using?

latest

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?
--

## Steps to reproduce the issue

1. use Cv-dropdown-item component
2. Run IBM Equal Access Accessibility Checker
3. Open violations and go to 4.1.2 Name, Role, Value
<img width="607" alt="Captura de Tela 2022-07-18 às 12 54 40" src="https://user-images.githubusercontent.com/25625728/179553191-e73e9891-b1c5-4f0e-bca3-ea357fcf881e.png">

4. etc.

## Additional information

- Screenshots or code
- Notes

PR #1347, opened 2022-07-19
fix: expanded not true even for child

barbarabeat:
if the expanded is not true the children will be hidden also

Contributes to [#1343](https://github.com/carbon-design-system/carbon-components-vue/issues/1343)

## What did you do?
"Hid" the children elements when some panel is not expanded.

## Why did you do it?

For screen readers, keyboards, these elements were still visible. So, it brings some accessibility problems.

## How have you tested it?

With 'IBM Equal Access Accessibility Checker ' and screen reader

- [ ] N/A
- [ x] No
- [ ] Yes

before:
Captura de Tela 2022-07-19 às 13 09 15

after:
Captura de Tela 2022-07-19 às 13 07 18

### DIFF
--- a/packages/core/src/components/cv-ui-shell/cv-header.vue
+++ b/packages/core/src/components/cv-ui-shell/cv-header.vue
@@ -79,7 +79,16 @@ export default {
         for (let index in this.panelControllers) {
           this.panelControllers[index].panelExpanded = false;
         }
-
+        const componentChildren = document.getElementsByClassName('cv-header-panel bx--header-panel');
+        if (newValue === false) {
+          for (let index = 0; index < componentChildren.length; index++) {
+            componentChildren[index].hidden = true;
+          }
+        } else {
+          for (let index = 0; index < componentChildren.length; index++) {
+            componentChildren[index].hidden = false;
+          }
+        }
         srcComponent.panelExpanded = newValue;
         this.panels[foundIndex].panelExpanded = newValue;
       }


lee-chase (dismissed):

davidnixon:
I don't know what @lee-chase  thinks but the query is concerning: `const componentChildren = document.getElementsByClassName('cv-header-panel bx--header-panel');` The `bx` part at least should honor the `carbonPrefix` that is used in other parts of that file.
Also the query starts with document instead of maybe something scoped to the component.

The biggest problem is that the left/right panels are just slots meaning they can hold anything. Notice there are no violations from the left slot because it is empty.

For example, here is a [sandbox example](https://codesandbox.io/s/carbon-components-vue-cvheader-sl3rsc?file=/src/TestComponent.vue:1863-1878) with a in-accessible link in the left panel.

It seems this fix is better suited to the `cv-header-panel` component.  You could probably add a ref to the slot there and enumerate the children from there. Also can I ask why `hidden = true;` instead of `ariaHidden = true`?

davidnixon (changes requested):
see comments - I don't think this should be merge for this component

lee-chase:
Sorry, yes just dismissed my own review which looked either in error or in haste.

On mounting the panel raises 'cv:panel-mounted' to the header. The panel should hide it's children. Probably in reaction to having panelExpanded, in the same function, change to false. There is an edge case I suppose where the panel header could cause it to bounce from true, to false to true again.

Probably as part o fthis setter

image

barbarabeat:
@davidnixon , @lee-chase , sorry for the delay...

About how to take the element, do you mean to use "this.$children" instead of 'cv-header-panel bx--header-panel'? I'm not sure if I understood... ''cv-header-panel bx--header-panel'' is how the buttons appear... I mean, it doesn't have another class together...

About hidden/aria-hidden, aria-hidden is not being enough.  Also, I was thinking with the idea of ["Content that was previously needed but is not any longer"](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden)..

At least, using ' :hidden="!panelExpanded ? 'true' : 'false'"' in CvHeaderPanel breaks the component. The panel doesn't open more...

lee-chase:
Hey @barbarabeat all children are registered with the CvHeader listening for 'cv:panel-mounted'. In functions like onCvPanelControlToggle you can see a panel being updated. This is preferable to using a query selector.

Issue #1350, opened 2022-07-20
Accessibility issue with CvOverflowMenu component


caroluchoa:
## Detailed description
This issue is related to CvOverflowMenu. Using IBM Accessibility Checker on the CvOverflowMenu results on the following violation: "Form control element <button> has no associated label. Violation"
I have uploaded a print of accessibility checker scan showing the violation and the code.

## Additional information

- Screenshots or code

<img width="685" alt="image" src="https://user-images.githubusercontent.com/43659990/180068842-a379f511-855c-4786-8a0b-abb2dcbd47a2.png">

Piece of code that caused the violation:

```html
<cv-overflow-menu id="header-overflow" :flip-menu="true" aria-label="menu">
       <cv-overflow-menu-item @click="applicationHistory" aria-label="History" </cv-overflow-menu-item>
       <cv-overflow-menu-item @click="editApplication" aria-label="Edit application">
           Edit application
       </cv-overflow-menu-item>
       <cv-overflow-menu-item @click="deleteApplication" aria-label="Delete application">
           Delete application
       </cv-overflow-menu-item>
</cv-overflow-menu>
```

PR #1351, opened 2022-07-20
fix: add aria-labelledby tag to CvOverflowMenu child button

caroluchoa:
Contributes to #1350

## What did you do?

Add aria-labelledby tag to CvOverflowMenu child button. Changed it to be labeled by the parent div.

## Why did you do it?

Child button was causing a label accessibility violation

## How have you tested it?
Using Accssebility Checker tool

- [X] N/A
- [ ] No
- [ ] Yes


packages/core/src/components/cv-overflow-menu/cv-overflow-menu.vue
```diff
@@ -14,6 +14,7 @@
       type="button"
       :aria-expanded="open ? 'true' : 'false'"
       :aria-controls="`${uid}-menu`"
+      :aria-labelledby="`${uid}-menu`"
```
caroluchoa, line 17:
```suggestion
      :aria-labelledby="`${uid}`"
```

caroluchoa:
@lee-chase Could you take a look on that one? thanks in advance Lee!

lee-chase (approved):

Issue #1355, opened 2022-08-11
Accessibility violation on CvFileUploader component


caroluchoa:
## Detailed description

> Describe in detail the issue you're having.
This issue is related to CvFileUploader. Using IBM Accessibility Checker on the CvFileUploader results on the following violation: **"The tabbable element's role 'none' is not a widget role"**
I have uploaded a print of accessibility checker scan showing the violation and the code.

> Is this a feature request (new component, new icon), a bug, or a general issue?
Bug

> Is this issue related to a specific component?
CvFileUploader

> What did you expect to happen? What happened instead? What would you like to see changed?
Was expecting no violations. Got an accessibility violation as showed on the screenshot below.

> What browser are you working in?
Chrome

> What version of the Carbon Design System are you using?
"@carbon/vue": "^2.44.0"

## Additional information

- Screenshots or code
![image](https://user-images.githubusercontent.com/43659990/184219399-250ff7e0-d959-4fd1-9eb4-ff2d1275808a.png)

```html
  <cv-file-uploader
       ref="fileUploader"
       v-model="receiversFile"
       data-cy="fileUploader"
       class="email-drag-drop"
       kind="drag-target"
       accept=".xlsx"
       :clear-on-reselect="true"
       :removable="true">
 </cv-file-uploader>
```

- Notes

PR #1356, opened 2022-08-11
fix: fix accessibility violation from CvFileUploader

caroluchoa:
Contributes to #1355

## What did you do?
I've removed the **tabindex** from the label tag to the div from the cv-wrapper and added a button role to that div to remove role violations.

## Why did you do it?
I did that to remove the violation reported on issue #1355. I can not just add a widget role to the label tag because it is label. So in that way, I've moved all the focus/role/tabindex to the child div.

## How have you tested it?
I've ran IBM accessibility checker and the violation was gone, with none extra violations added.

- [X] N/A
- [ ] No
- [ ] Yes


packages/core/src/components/cv-file-uploader/cv-file-uploader.vue
```diff
@@ -25,6 +21,10 @@
           @dragover="onDragEvent"
           @dragleave="onDragEvent"
           @drop="onDragEvent"
+          tabindex="0"
```
caroluchoa, line 24:
```suggestion
          role="button"
          tabindex="0"
```

lee-chase:
Unfortunately, this breaks the Carbon focus styling. It might be possible to rectify by moving the styling class down one level but this would need further verification. 

Note this component is based on https://v10-react.carbondesignsystem.com/?path=/story/components-fileuploader--drag-and-drop-upload-container-example-application

To proceed you must demonstrate the component functions in the same way and that no styling is broken.

caroluchoa:
> Unfortunately, this breaks the Carbon focus styling. It might be possible to rectify by moving the styling class down one level but this would need further verification.
> 
> Note this component is based on https://v10-react.carbondesignsystem.com/?path=/story/components-fileuploader--drag-and-drop-upload-container-example-application
> 
> To proceed you must demonstrate the component functions in the same way and that no styling is broken.

Ok! Thanks Lee! I'm gonna try that out

caroluchoa:
> Unfortunately, this breaks the Carbon focus styling. It might be possible to rectify by moving the styling class down one level but this would need further verification.
> 
> Note this component is based on https://v10-react.carbondesignsystem.com/?path=/story/components-fileuploader--drag-and-drop-upload-container-example-application
> 
> To proceed you must demonstrate the component functions in the same way and that no styling is broken.

Hi Lee, I have moved the label tag styling to CvWrapper component to avoid breaking focus styling you mentioned. There was the need of creating a computed property with the classname strings due to the fact that CvWrapper components does not accept Vue dynamic class binding, because the class property is created like:

```javascript
return createElement(
        context.props.tagType,
        {
          attrs: context.data.attrs,
          class: `${context.data.class || ''} ${context.data.staticClass || ''}`.trim(),
          style: { ...context.data.staticStyle, ...context.data.style },
          on: context.listeners,
        },
```
While dynamic class binding sends an object (with classes names as keys and a boolean as value) to context.data.class

PR #1361, opened 2022-08-31
docs: index compatible design kits

mattrosno:
Related: https://github.com/carbon-design-system/carbon-platform/issues/967

The Carbon Vue library has a design kits page, to list compatible design kits with the components/patterns in the Carbon Vue library: https://next.carbondesignsystem.com/libraries/carbon-components-vue/latest/design-kits

This list of `designKits` would normally be inherited from the `carbon-styles` library (PR https://github.com/carbon-design-system/carbon/pull/12033), but Carbon Vue has yet to upgrade to v11.

By not inheriting `designKits`, we can specify in this library so the v10 Figma kits are listed, not the new v11 Figma kits.

Once Carbon Vue updates to v11, we can remove the `designKits` array in the config file.

## What did you do?

- Added `designKits` array to the index

## Why did you do it?

- See above

## How have you tested it?

- The YML is valid according to our schema
- We're building out the https://next.carbondesignsystem.com/libraries/carbon-components-vue/latest/design-kits page now, so we'll catch any errors should they exist

- [x] N/A
- [ ] No
- [ ] Yes

### DIFF
--- a/packages/core/src/carbon.yml
+++ b/packages/core/src/carbon.yml
@@ -23,6 +23,41 @@ library:
           path: 'https://github.com/carbon-design-system/carbon-website/blob/carbon-platform/src/pages/developing/vue-tutorial/step-5.mdx'
         - title: Wrapping up
           path: 'https://github.com/carbon-design-system/carbon-website/blob/main/src/pages/developing/vue-tutorial/wrapping-up.mdx'
+  designKits:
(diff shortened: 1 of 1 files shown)


alisonjoseph (approved):

aubrey-oneal (approved):
🎉 looks good

Issue #1387, opened 2023-01-03
CvOverflowMenu does not have aria-label props


lbenicio-ibm:
## Detailed description

The component Cv Overflow Menu, which create an overflow menu with a button as trigger, does not support appropriated aria-label props on that button.

> Describe in detail the issue you're having.

I am having difficulties in creating accessibility features for screen readers.

> Is this a feature request (new component, new icon), a bug, or a general issue?

Bug

> Is this issue related to a specific component?

Cv Overflow Menu

> What did you expect to happen? What happened instead? What would you like to see changed? 

I expected to pass prop "aria-label" to Cv Overflow Menu components, and it properly adds this prop to its button.

> What browser are you working in?

Safari

> What version of the Carbon Design System are you using?

10.46.0

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

No

## Steps to reproduce the issue

1. Create a Cv-Overflow-Menu component
2. Add "aria-label" prop
3. Inspect element on generated html in browser
4. generated button does not show proper aria-label

## Additional information

- Screenshots or code
- Notes

PR #1388, opened 2023-01-03
feat(cvOverflowMenu): add aria-label support

lbenicio-ibm:
Contributes to #1387

## What did you do?

Added support to aria-label prop CvOverflowMenu component

## Why did you do it?

Need to support screen readers on our application

## How have you tested it?

i have mounted the component local on my machine

- [ ] N/A
- [ ] No
- [x] Yes


lee-chase:
Is there an accessibility issue reported by the Equal Accessibility Checker I can check before and after the fix?

What are the other prop values you are using?
Are you using the `label` prop?

lbenicio-ibm:
> Is there an accessibility issue reported by the Equal Accessibility Checker I can check before and after the fix?
> 
> What are the other prop values you are using? Are you using the `label` prop?

1. Yes, there are accessibility reports. Line 36 (shown below) on before shows the error I am trying to fix: 

```
Vue App | http://localhost:8090/notification | scan9 | 4183703059 | Violation | 1 | 4.1.2 Name, Role, Value | A | WCAG20_Input_ExplicitLabel | Form control element <button> has no associated label | button | <button class="bx--overflow-menu__trigger bx--tooltip__trigger bx--tooltip--a11y" id="filter-overflow-trigger" aria-controls="filter-overflow-menu" aria-expanded="false" type="button" aria-haspopup=""> | /document[1]/main[1]/button[2] | https://able.ibm.com/rules/tools/help/WCAG20_Input_ExplicitLabel |   |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
```


[before.xlsx](https://github.com/carbon-design-system/carbon-components-vue/files/10352500/before.xlsx)
[after.xlsx](https://github.com/carbon-design-system/carbon-components-vue/files/10352501/after.xlsx)


here it is the original code which uses the component:

```javascript
<cv-overflow-menu id="filter-overflow" :flip-menu="true" class="bx--toolbar-action" aria-label="Filter menu">
            <template slot="trigger">
              <Filter16 class="bx--overflow-menu__icon bx--toolbar-filter-icon" />
            </template>
            <div class="bx--row filter-content">
              <div class="bx--column">
                <p style="font-size: 0.75rem">Notification type</p>
                <cv-checkbox
                  v-for="kind in notificationKind"
                  :key="kind"
                  v-model="checks"
                  :value="kind"
                  :label="kind"
                />
              </div>
              <div style="padding-left: 2rem" class="bx--column">
                <p style="font-size: 0.75rem">Notifications from</p>
                <cv-toggle
                  :checked="checked"
                  :small="true"
                  :value="toggle"
                  aria-label="All applications"
                  @change="toggleChanged"
                >
                  <template slot="text-left">All applications</template>
                  <template slot="text-right">All applications</template>
                </cv-toggle>
              </div>
            </div>
          </cv-overflow-menu>
```

lee-chase:
@lbenicio-ibm the button has a `label` are you using it?

lbenicio-ibm:
@lee-chase I tried with the label, but it did not add the aria-label property while using label

lee-chase:
@lbenicio-ibm no but I would expect the `aria-labelledby` to negate the need for a separate aria-label.

lee-chase (approved):

PR #1392, opened 2023-01-06
feat: using native Crypto API to generate UUIDs

edo9k:
## What did you do?

Replaced an obscure UUID generation function with a native method from the Crypto API (available everywhere). 

About the Crypto API and its availability: https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

## Why did you do it? 

Using the native API was about 50% faster, and it's all done in a single line of code, calling the Crypto API.
The current version is somewhat obscure. 

The in-browser benchmark can run here: https://jsbench.me/a9lcjwwi2p/2

## How have you tested it?

Function output was unaltered with the change. Both versions output `uid-${ a_random_uuid }`

- [x] N/A
- [ ] No
- [ ] Yes

### DIFF
--- a/packages/core/src/mixins/uid-mixin.js
+++ b/packages/core/src/mixins/uid-mixin.js
@@ -1,12 +1,6 @@
 const idGen = {
   newId() {
-    // as it is funny gen IDs using b - https://gist.github.com/jed/982883
-    const b = a =>
-      a
-        ? (a ^ ((Math.random() * 16) >> (a / 4))).toString(16)
-        : ([1e7] + -1e3 + -4e3 + -8e3 + -1e11).replace(/[018]/g, b);
-
-    return `uid-${b()}`;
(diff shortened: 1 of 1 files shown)


lee-chase:
This would likely break server-side rendering.

Elsewhere we use the following based on // https://stackoverflow.com/a/2117523

```
function uuidv4() {
  const randomValues = (c) =>
    typeof crypto !== 'undefined'
      ? crypto.getRandomValues(new Uint8Array(1))[0] & (15 >> (c / 4))
      : (Math.random() * 16) >> (c / 4);

  return ([1e7] + -1e3 + -4e3 + -8e3 + -1e11).replace(/[018]/g, (c) =>
    (c ^ randomValues(c)).toString(16)
  );
}
```

lee-chase (changes requested):
See comment regarding server side rendering

PR #1393, opened 2023-01-06
fix(cv-data-table): Fix accessibility violations from CvDataTable

bmamprim:
Contributes to #1394

## What did you do?
I added role and aria-label attributes to remove accessibility violations

## Why did you do it?
I did that to remove violations reported on issue #1394

## How have you tested it?
I've ran IBM accessibility checker and the violations was gone, with none extra violations added.

- [x] N/A
- [ ] No
- [ ] Yes


lee-chase:
@bmamprim cannot use hard-coded aria-label. It should either be labelled by `:id="`${uid}-search`"` or use a property for the label.

lee-chase (changes requested):
No fixed strings.

Add a property and set a default if needed.

PR #1420, opened 2023-02-14
fix(cv-datatable): batch action scroll on select/unselect

peterzaccha:
Contributes to #

https://vue.carbondesignsystem.com/?path=/story/components-cvdatatable--infinite-pagination

This PR fixes a bug in the datatable component.

Using the batch actions in the datatable,
if you scroll past the top of the table and select/unselect a row, you will be teleported back to the top,

this behavior is because the checkbox input has an absolute position, and the nearest container with a relative position is the outer container for the table

## What did you do?

added a style to the container of the checkbox, to make its position relative

## Why did you do it?

to limit the checkbox input which has an absolute position from moving away to the closer relative container,

- [ ] N/A
- [ ] No
- [ ] Yes

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -5,32 +5,21 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 
 ## [2.45.1](https://github.com/carbon-design-system/carbon-components-vue/compare/@carbon/vue@2.45.0...@carbon/vue@2.45.1) (2023-01-24)
 
-
 ### Bug Fixes
 
-* set aria role seperator on modal before/after content ([#1409](https://github.com/carbon-design-system/carbon-components-vue/issues/1409)) ([21b3756](https://github.com/carbon-design-system/carbon-components-vue/commit/21b375650de2e020ff96c5e3810e3393bbd318a8))
-
-
-
-
+- set aria role seperator on modal before/after content ([#1409](https://github.com/carbon-design-system/carbon-components-vue/issues/1409)) ([21b3756](https://github.com/carbon-design-system/carbon-components-vue/commit/21b375650de2e020ff96c5e3810e3393bbd318a8))
 
 # [2.45.0](https://github.com/carbon-design-system/carbon-components-vue/compare/@carbon/vue@2.44.2...@carbon/vue@2.45.0) (2023-01-10)
 
-
 ### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


edo9k (approved):
This is a small CSS change that fixes a very annoying problem. 

It prevents the UI from jumping every time the user changes the status of a checkbox.

packages/core/CHANGELOG.md
```diff
@@ -5,32 +5,21 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 
 ## [2.45.1](https://github.com/carbon-design-system/carbon-components-vue/compare/@carbon/vue@2.45.0...@carbon/vue@2.45.1) (2023-01-24)
 
-
```
davidnixon, line 8: Can you revert the changes to this file. I think they are unintentional.
```sh
git checkout 615311b8faab77997f768f38acee3d9a2a5e57dc packages/core/CHANGELOG.md
git add packages/core/CHANGELOG.md
git commit
```

davidnixon (changes requested):

packages/core/src/components/cv-checkbox/cv-checkbox.vue
```diff
@@ -1,8 +1,10 @@
 <template>
   <div
-    style="{position: 'relative'}"
     class="cv-checkbox"
-    :class="[`${carbonPrefix}--checkbox-wrapper`, { [`${carbonPrefix}--form-item`]: formItem }]"
+    :class="[
+      `${carbonPrefix}--checkbox-${inline ? 'inline' : 'wrapper'}`,
```
davidnixon: should be `--checkbox--` instead of `--checkbox-`

davidnixon (changes requested):

packages/core/src/components/cv-checkbox/cv-checkbox.vue
peterzaccha: the wrapper class is `--checkbox-wrapper` while the inline one is `--checkbox--inline` (with double dashes)

packages/core/CHANGELOG.md
peterzaccha: [77eb4c8](https://github.com/carbon-design-system/carbon-components-vue/pull/1420/commits/77eb4c892996e3cd0c3d0160ce7522df461d7008)

packages/core/src/components/cv-checkbox/cv-checkbox.vue
davidnixon: nice

davidnixon (approved):

davidnixon:
@lee-chase OK to merge this one? I ran the storybook locally to verify the fix.

PR #1422, opened 2023-02-27
fix(cv-data-table): accessibility violation - batch action cancel button

peterzaccha:
Contributes to #

This PR fixes the accessibility violation 
Element "button" should not be focusable within the subtree of an element with an 'aria-hidden' attribute with value 'true'

## What did you do?

inspired by the react carbon lib
i check if the batch actions are not active, set the tabindex on the cancel button to -1 

- [x] N/A
- [ ] No
- [ ] Yes

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -5,21 +5,32 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 
 ## [2.45.1](https://github.com/carbon-design-system/carbon-components-vue/compare/@carbon/vue@2.45.0...@carbon/vue@2.45.1) (2023-01-24)
 
+
 ### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


davidnixon (changes requested):
:-( Sorry for being so slow.
1. I think this is a good fix and thanks for taking the time to work on it.
2. Looking at the [React storybook](https://v7-react.carbondesignsystem.com/?path=/story/components-datatable-batch-actions--default) the tab index is toggled when the batch action is active just as you have done here
3. However in react all the buttons under the batch action are toggled

Could you tweak this to watch the `batchActive` data and react to changes by find and toggle all the buttons under batch actions?

I think you will need a ref on the `--action-list` div and a query like:
```js
toggle(val) {
buttons=this.actionList.querySelectorAll('button')
buttons.forEach((b) => b.tabIndex = val)
}
```

PR #1424, opened 2023-03-01
fix: set immediate to true of expanded watch prop from side nav menu

dvpfran:
Contributes to #

## What did you do?
Basically I set the prop `immediate` to `true` of `expanded` watch

## Why did you do it?
By default, the prop `immediate` is `false` and when we try to set to `true` it's ignored
```vue
<CvSideNavMenu :title="title" :expanded="true">
...
</CvSideNavMenu>
``` 

## How have you tested it?
I forked the repo and make the changes. Then I tested in my app and it worked.

- [x] N/A
- [X] No
- [x] Yes

### DIFF
--- a/src/components/CvUIShell/CvSideNavMenu.vue
+++ b/src/components/CvUIShell/CvSideNavMenu.vue
@@ -79,6 +79,7 @@ watch(
   () => props.expanded,
   value => {
     data.isExpanded = value;
-  }
+  },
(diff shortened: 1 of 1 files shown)


dvpfran:
@davidnixon, sorry to ping you but can you have a look on this?

davidnixon (approved):

PR #1426, opened 2023-03-09
fix(cv-combo-box): accessibility violations

peterzaccha:
Contributes to #

## What did you do?

Fix accessibility violation: nested-interactive
caused by using a listbox role on a div with two buttons

this fix inspired by the same component in the react carbon lib
https://react.carbondesignsystem.com/?path=/story/components-combobox--default

- [ ] N/A
- [ ] No
- [ ] Yes


davidnixon (changes requested):
This looks good to me. It matches what in [react storybook](https://v7-react.carbondesignsystem.com/?path=/story/components-combobox--playground)

Do you have a before & after screenshot of the accessibility violations?

PR #1432, opened 2023-03-13
chore: add workflow for 'next' package

davidnixon:
## What did you do?

Added workflow to publish "next" package

## How have you tested it?

Tested locally using local nm repo verdaccio
- Published both Vue 2 package and Vue 3 next package to local npm repo ` yarn lerna publish patch --no-git-tag-version --no-push --yes --registry http://localhost:4873/`
- Verified that installing via "npm install @carbon/vue" installs the Vue 2 package
- Verified that installing via "npm install @carbon/vue@next" installs the Vue 3 package "3.0.3-alpha.0"

- [ ] N/A


lee-chase (changes requested):
Only question would by why `prepatch` instead of `--create-release` but guessing you have tested it.

Might want to just update the dry run comments above to match.

davidnixon:
- The `prepatch` does 2 things. The "patch" part  increments the last digit so that `3.0.3-alpha.0` becomes `3.0.4-alpha.0` . The "pre" part adds "alpha" string. 
- Removing the `--create-release` means that a new GitHub release is NOT created. Which I think is the right thing  for now. A GitHub tag is created but its not marked as a new release.

I did test everything locally but its difficult with GitHub actions to test everything end to end. I test all the steps in the action locally.

lee-chase (approved):

Issue #1434, opened 2023-03-19
CvLink work in progress


felipebritor:
Port link component to vue3
https://vue.carbondesignsystem.com/?path=/story/components-cvlink--a

felipebritor:
Hello, @davidnixon @lee-chase. 

I ended up already opening a PR for this, but I have a question. There are a few changes in the react implementation not covered for vue2:
- A place to pass an icon matching the link specification (right side of the text with 16px);
- A 'visited' prop to set the link as visited;
- Sizing.

Should I also add it to the PR? I basically replicated what was done for vue2

felipebritor:
Thanks, David. I'll update the PR later today

PR #1435, opened 2023-03-20
feat: CvLink implementation for Vue3

felipebritor:
Contributes to #1434 

## What did you do?
Port CvLink component to vue3, covering the same functionalities at the vue2 implementation and a small update in 'useLinkProps' to also add 'role' attribute when link is disabled.

Edit: I've also left the story more like the react one: a child story for each 'state' instead of how it's in vue2.

## How have you tested it?
Yes. Although, Since most tests assert template changes based on props received, they are not exactly following `@testing-library`'s guidelines on replicating user behavior. 

- [ ] N/A
- [ ] No
- [x] Yes

### DIFF
--- a/src/components/CvLink/CvLink.stories.js
+++ b/src/components/CvLink/CvLink.stories.js
@@ -5,7 +5,7 @@ import {
 } from '../../global/storybook-utils';
 
 import { CvLink } from '.';
-import { props as propsCvLink } from '../../use/cvLink';
+import { props as propsCvLink, linkSizes } from '../../use/cvLink';
 
 export default {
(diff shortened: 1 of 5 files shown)


davidnixon:
:fireworks: Thanks for this contribution! I should be able to review it tomorrow. Sorry for the delay.

davidnixon:
@felipebritor LGTM just holding this one because I think you are adding some changes from React per discussion in #1434

felipebritor:
@davidnixon PR updated with the items mentioned at #1434 👍

davidnixon (approved):
Awesome. LGTM

PR #1436, opened 2023-03-20
fix: sb vue2 & vue3 paths

davidnixon:
## What did you do?

Changed the links in the vue3 vue2 storybooks to properly point to the vue2/vue3 versions. 

## Why did you do it?

The combined storybook build and deployed worked great. The links:
- vue2 is here https://vue.carbondesignsystem.com/?path=/story/welcome--default
- vue3 is here https://vue.carbondesignsystem.com/vue3/?path=/story/carbon--welcome

However the links in the storkbooks to vue3 are broken. ☹️ 

The current link to vue3 is `https://vue.carbondesignsystem.com/carbon-components-vue/vue3/`
Notice the extra **carbon-components-vue** in the path. This is because when testing with GitHub pages, the repo name is part of the path but when accessed via the custom domain, its not there. 

I added a compute property to calculate the vue2 & vue3 URLs based on if `carbon-components-vue` is in the URL path or not. When building and testing locally it is not there and so the path to vue3 is just "/vue3". When building an testing in my own repo the "carbon-components-vue" is part of the path and so the path to vue3 is "/carbon-components-vue/vue3/"
https://davidnixon.github.io/carbon-components-vue/?path=/story/welcome--default
https://davidnixon.github.io/carbon-components-vue/vue3/?path=/story/carbon--welcome

## How have you tested it?

Locally and in my own GitHub pages deploy.

- [x] N/A

### DIFF
--- a/storybook/_storybook/views/sv-template-view/sv-versions.vue
+++ b/storybook/_storybook/views/sv-template-view/sv-versions.vue
@@ -15,7 +15,7 @@
     <div class="versions__info">
       <div class="bx--row">
         <div class="bx--no-gutter-md--left bx--col-md-4 bx--col-lg-4">
-          <a class="versions__card" href="/carbon-components-vue/vue3/">
+          <a class="versions__card" :href="vue3Url">
             <div class="bx--aspect-ratio bx--aspect-ratio--2x1">
               <div class="bx--aspect-ratio--object">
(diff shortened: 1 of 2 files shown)


davidnixon:
Conformed fix in Safari & FF

PR #1442, opened 2023-04-01
fix: datepicker issue for mobile

kirangurung2023:
Contributes to #

## What did you do?
I created a pull request for the current date time picker component library, modifying the datetime picker to have the option disableMobile: true. This change disables the automatic detection of mobile browsers by the flatpickr library and prevents the issue of turning the date input into a native date/time/datetime input when it is not desired.

## Why did you do it?
The flatpickr library has a known issue, as reported in their GitHub repository (https://github.com/flatpickr/flatpickr/issues/1115), where the automatic detection of mobile browsers may not work as intended. This temporary fix allows users to avoid the problem until a more permanent solution is implemented by the flatpickr team.

## How have you tested it?
I tested the changes on various devices and browsers, including mobile and desktop environments, to ensure that the datetime picker now behaves as expected and does not automatically convert the date input to a native input type on mobile browsers.

- [x] N/A
- [ ] No
- [ ] Yes

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -5,32 +5,21 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 
 ## [2.45.1](https://github.com/carbon-design-system/carbon-components-vue/compare/@carbon/vue@2.45.0...@carbon/vue@2.45.1) (2023-01-24)
 
-
 ### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


davidnixon:
I think you can already set this and other options in the `calOptions` property which is then passed to the flatpicker.

Could you try that and let me know if there is still an issue. Maybe an update to the storybook to alert users to the issue?

PR #1443, opened 2023-04-03
fix: add yarn cache to .gitignore and install @mdx-js/react

grievous0:
Contributes to #

## What did you do?
Added yarn/cache to the .gitignore file and installed the @mdx-js/react dependency by adding it to the package.json file.

## Why did you do it?
The yarn/cache directory was added to the git repository and it should be ignored. Also, the @mdx-js/react dependency was required by the Storybook but was not installed previously.

## How have you tested it?
I tested the changes locally by running the Storybook and verifying that it worked correctly. I also verified that the yarn/cache directory was ignored by Git after the changes were made to the .gitignore file.

- [x] N/A
- [ ] No
- [ ] Yes


grievous0:
Hey, @davidnixon! Could you review & approve or reject these changes?

davidnixon:
@lee-chase Do you have an opinion the yarn change?
I'm using the yarn Q&A here https://yarnpkg.com/getting-started/qa when setting this file originally.

@grievous0 checking I'm looking at the mdx change now. Do you get a build error from Storybook without it?

grievous0:
@davidnixon Yes! I get an error and I can't run Storybook without this dependency, I asked some peers to test it for me and they got the same error!

davidnixon:
This is what I see from `yarn why @mdx-js/react`
```
├─ @storybook/addon-docs@npm:6.5.15
│  └─ @mdx-js/react@npm:1.6.22 (via npm:^1.6.22)
│
└─ @storybook/addon-docs@npm:6.5.15 [c2f80]
   └─ @mdx-js/react@npm:1.6.22 [8cb1c] (via npm:^1.6.22 [8cb1c])
```
Meaning its already installed because its a dependency from ` @storybook/addon-docs`

`yarn why @storybook/addon-docs`
```
├─ @storybook/addon-essentials@npm:6.5.15
│  └─ @storybook/addon-docs@npm:6.5.15 (via npm:6.5.15)
│
└─ @storybook/addon-essentials@npm:6.5.15 [51748]
   └─ @storybook/addon-docs@npm:6.5.15 [c2f80] (via npm:6.5.15 [c2f80])
```

But `@storybook/addon-docs` is not explicitly listed in the package.json so I think this may be the root issue.

Can you undo this change and try adding  `@storybook/addon-docs` instead?

Based on this doc https://yarnpkg.com/getting-started/qa I think you should undo the .gitignore change

grievous0:
> Isso é o que eu vejo de`yarn why @mdx-js/react`
> 
> ```
> ├─ @storybook/addon-docs@npm:6.5.15
> │  └─ @mdx-js/react@npm:1.6.22 (via npm:^1.6.22)
> │
> └─ @storybook/addon-docs@npm:6.5.15 [c2f80]
>    └─ @mdx-js/react@npm:1.6.22 [8cb1c] (via npm:^1.6.22 [8cb1c])
> ```
> 
> Significa que já está instalado porque é uma dependência de` @storybook/addon-docs`
> 
> `yarn why @storybook/addon-docs`
> 
> ```
> ├─ @storybook/addon-essentials@npm:6.5.15
> │  └─ @storybook/addon-docs@npm:6.5.15 (via npm:6.5.15)
> │
> └─ @storybook/addon-essentials@npm:6.5.15 [51748]
>    └─ @storybook/addon-docs@npm:6.5.15 [c2f80] (via npm:6.5.15 [c2f80])
> ```
> 
> Mas `@storybook/addon-docs`não está listado explicitamente no package.json, então acho que esse pode ser o problema raiz.
> 
> Você pode desfazer essa alteração e tentar adicionar `@storybook/addon-docs`em seu lugar?
> 
> Com base neste documento https://yarnpkg.com/getting-started/qa , acho que você deve desfazer a alteração do .gitignore

All right! Changes made, could you review them again?

davidnixon:
I think I understand the original issue a bit more now. When I look at the yarn.lock file in the PR I see this:
```
# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
# yarn lockfile v1


"@achrinza/node-ipc@^9.2.5":
```

But the current lock file on the vNext branch looks like this:
```
# This file is generated by running "yarn install" inside your project.
# Manual changes might be lost - proceed with caution!

__metadata:
  version: 6
  cacheKey: 8

"@achrinza/node-ipc@npm:^9.2.5":
```

Your lock file is generated by yarn v1 but the vNext branch is using yarn v3. I do not understand why that happened. The yarn command should see the `.yarnrc.yml` in the repo and use yarn v3.

I think you should re-clone the repo into a new folder and redo the change there. Also, we are not quite sing storybook 7 yet so the addon should be at the 6 level.
```
yarn add @storybook/addon-docs@6
```

I think the yarn v1 issue is also how you were able to find this bug in the first place.  I think yarn v3 was installing the `@storybook/addon-docs` even though it was incorrectly omitted in the package.json.

davidnixon (changes requested):
See comments. Thank you for working on this!

grievous0:
> Ver comentários. Obrigado por trabalhar nisso!

No problem! Changes made, I re-clone the repo and fix the storybook version, yarn.lock looks good now!

grievous0:
> See comments. Thank you for working on this!

No problems! Changes made, I re-clone the repo and fix the storybook version, yarn.lock looks good now!

davidnixon:
Still some strangeness. :-)
image

Checking to see if I can fix on my end ...

davidnixon:
Thanks for you perseverance on this one. I think you can fix like this assuming that `upstream` is this repo and `origin` is your fork.
1. Fetch all changes in this repo
1. switch to upstream (this repo) vNext
2. checkout a new branch in origin (your repo)
3. get the package.json file from your last change
4. run yarn install to update the lock file 
5. commit the changes which should **only be about 24 files**
6. overwrite this PR with content from the new branch
```sh
git fetch --all
git checkout upstream/vNext
git switch -c chore-fix-package
git checkout f15bc8bb3ae9cc23b412775dcfef170b6fbc1e2f -- storybook/package.json
yarn
git add yarn.lock .yarn/cache
git push --force -u origin chore-fix-package:update/gitignore-and-packagejson
```

That last bit with the "--force" will throwaway everything on this PR except the changes from the new branch `chore-fix-package`

davidnixon (approved):
lgtm

Issue #1437, opened 2023-03-21
CvTextInput work in progress


felipebritor:
Port text input, also covering password, components to vue3
https://vue.carbondesignsystem.com/?path=/story/components-cvtextinput--default

---

Hey @davidnixon @lee-chase, I'm not able to add labels to my issue. This one and #1434 need the v3 one. May you help me?

PR #1445, opened 2023-04-13
feat: cv text & password input - vue3

felipebritor:
Contributes to #1437 

## What did you do?
CvTextInput (and 'CvPasswordInput') implementation for vue3, following vue2 implementation and react-carbon v7 as base

## How have you tested it?
Component features were added incrementally, per commit. Unit tests were added on each step

- [x] Yes

### DIFF
--- a/src/components/CvTextInput/CvTextInput.stories.js
+++ b/src/components/CvTextInput/CvTextInput.stories.js
@@ -160,6 +160,7 @@ export default {
     'update:modelValue': {
       type: 'event',
       table: {
+        type: { summary: 'event' },
         category: 'events',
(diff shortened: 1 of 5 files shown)


felipebritor:
Hard pushed 2 changes:
- Removed `$attrs || cvId` on label's for attr. `useCvId` already handle id prop [18b8dbf](https://github.com/carbon-design-system/carbon-components-vue/pull/1445/commits/18b8dbfb3251b55cc8cd9e4579c2f89c6902b1e1)
- Remove a commit setting prop `value`, which, if set, is the initial text-input value.

PR #1448, opened 2023-04-14
fix: update vNext README.md file to include storybook instructions

Ana-Luz:
Contributes to `vNext` README.md file

## What did you do?
Update of the README.md file, adding the instructions on how to run the storybook.

## Why did you do it?
Because these instructions can help new contributors test their components.

## How have you tested it?
By watching the README.md file in my fork.

- [x] N/A
- [ ] No
- [ ] Yes

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,3 +72,22 @@ If you just want to build an individual package you can limit the scope:
 `yarn build --scope storybook`
 
 To start the storybook in a local server use `yarn start`.
+
+## How to run the storybook
+
+:bulb: **In this case, it's best to run using `npm`!**
+
+Just follow the steps listed below and you will be able to run the storybook.
+
+1. After the checkout to the `vNext` branch, in order to install the dependencies run the command `npm i` on the root;
+2. Now, run the command `cd storybook` to enter the storybook folder, then again run the command `npm i` to install the dependencies inside the storybook folder;
+3. Finally, run the command `npm run serve` inside the storybook folder.
+
+In other words, these are the commands you're going to use in order of execution:
+
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,3 +72,20 @@ If you just want to build an individual package you can limit the scope:
...
+1. After the checkout to the `vNext` branch, in order to install the dependencies run the command `yarn install` on the root;
+2. Now, run the command `cd storybook` to enter the storybook folder, then again run the command `yarn install` to install the dependencies inside the storybook folder;
+3. Finally, run the command `yarn serve` inside the storybook folder.
+
+In other words, these are the commands you're going to use in order of execution:
+
+```
+yarn install
+cd storybook
+yarn install
```
davidnixon, line 89: I do not think we need to run yarn install in the storybook directory? Because it is listed in the root as a workspace I think we can skip yarn install in the directory.

https://github.com/carbon-design-system/carbon-components-vue/blob/bbb90311e3b6d9e4f5a2ff92cb794b0622e11ecb/package.json#L24-L26

Could you try without that step and see if it works for you.

davidnixon (changes requested):
Thank you so much for updating this!
I had a small question about the second yarn install.

README.md
Ana-Luz: Sure, I'll try it right away!

Ana-Luz: It worked! Thanks a lot for the tip! I'm making the change in the README file right now, and then I'll commit it. The change should appear soon!

davidnixon (approved):

Issue #1451, opened 2023-04-19
CvForm, CvFormGroup and CvFormItem work in progress


felipebritor:
Port CvForm, CvFormGroup and CvFormItem components to vue3
https://vue.carbondesignsystem.com/?path=/story/components-cvform--default

davidnixon:
merged

PR #1452, opened 2023-04-19
feat: implement CvForm, CvFormGroup and CvFormItem

felipebritor:
Contributes to #1451 

## What did you do?
CvForm, CvFormGroup and CvFormItem implementation for vue3, following vue2 implementation and react-carbon v7 as base

## How have you tested it?
Using both unit tests and through the stories

- [x] Yes

### DIFF
--- a/.storybook/styles.scss
+++ b/.storybook/styles.scss
@@ -52,6 +52,11 @@ $carbon--theme: $carbon--theme--white;
   padding: $layout-04;
   background-color: $ui-background;
   color: $text-01;
+
+  // set space between form items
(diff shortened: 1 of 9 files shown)


davidnixon (changes requested):
Nice work!
The only thing I see is that the "Submit" button in the storybook actually open a new page. Looking at the Vue 2 version, the storybook has:
```vue
<cv-form @submit.prevent="actionSubmit">
```
Could you add something like that?

davidnixon (approved):

Issue #1463, opened 2023-06-20
Work in progress cv-inline-loading


davidnixon:
**Please read** [Vue 3 guidance here](https://github.com/carbon-design-system/carbon-components-vue/tree/vNext#vue-3-guidance)

Port cv-inline-loading to Vue 3

Link to Vue 2 storybook:
https://vue.carbondesignsystem.com/?path=/story/components-cvinlineloading--default

You may want to reference the [Carbon 10 React components](https://v7-react.carbondesignsystem.com/?path=/story/getting-started--welcome)

felipebritor:
Sure. I'll check tomorrow night

davidnixon:
merged

PR #1464, opened 2023-06-22
feat: port inline loading to vue 3

davidnixon:
Contributes to #1463

- [x] Yes


felipebritor:
Everything looks good, @davidnixon 👍

Issue #1460, opened 2023-06-12
CvToggle work in progress


WesleyGazz:
port to Vue 3
https://vue.carbondesignsystem.com/?path=/story/components-cvtoggle--default

PR #1467, opened 2023-07-04
feat: implement CvToggle

WesleyGazz:
Contributes to [#1460](https://github.com/carbon-design-system/carbon-components-vue/issues/1460)

## What did you do?
Implemented CvToggle to Carbon Vue3 

## Why did you do it?
To help community development 

## How have you tested it?
I Replicated unit tests from Vue2 to Vu3 

- [ ] N/A
- [ ] No
- [x] Yes


src/components/CvToggle/CvToggle-Skeleton.vue
```diff
@@ -0,0 +1,29 @@
+<template>
```
davidnixon, line 1: This one is not working. BTW I don't think a skeleton toggle is a very common use case but it would be good to have it working for completeness. Here is what I see:
image

The react component looks OK
https://v7-react.carbondesignsystem.com/?path=/story/components-toggle--skeleton

Have a look there but they have the bx--skeleton on an input element.

davidnixon (changes requested):
Thank you for working on this! Everything looks great except the skeleton mode could you have a look at that one in the react storybook and make changes here.

Thanks!

src/components/CvToggle/CvToggle-Skeleton.vue
WesleyGazz: Hello! In the vue2 the [Skeleton](https://vue.carbondesignsystem.com/?path=/story/components-cvtoggle--skeleton) component is the same, I created based on the current V2 version.

Do you have any special requirements for that?
image

davidnixon:
@WesleyGazz you are right. This is broken in Vue2 as well. https://vue.carbondesignsystem.com/?path=/story/components-cvtoggle--skeleton

Maybe we should merge this and just open an issue for Vue 2 & 3 regarding the skeleton. What do you think?

WesleyGazz:
@davidnixon I think it's a good idea, ok for me. Thanks

davidnixon (approved):

PR #1470, opened 2023-07-05
feat: Port CvButtonSkeleton to Vue3

OlkaB:
Contributes to #1469

## What did you do?
Port CvButtonSkeleton to Vue3

## Why did you do it?
Hope this will help to finish transition before Vue2 EOL

## How have you tested it?
please see CvButtonSkeleton.spec.js + CvButtonSkeleton.stories.js
compared behaviours with Vue2 storybook

## Were docs updated if needed?
- [x] Yes

PR #1475, opened 2023-07-12
fix(cv-button): fix CvButton stories, where buttons are blank (label …

OlkaB:
…is not shown)

Contributes to RFC from PR #1470

## What did you do?
Fixed stories with buttons without labels

## Why did you do it?
requested in PR #1470

## How have you tested it?
Visually in Storybook
image

- [x] N/A
- [ ] No
- [ ] Yes


davidnixon (approved):
Thanks!

Issue #1495, opened 2023-07-29
Work in progress cv-search


davidnixon:
**Please read** [Vue 3 guidance here](https://github.com/carbon-design-system/carbon-components-vue/tree/vNext#vue-3-guidance)

Port cv-searchto Vue 3

Vue 2 storybook here. For example:
https://vue.carbondesignsystem.com/?path=/story/components-cvsearch--default

Reference the [Carbon 10 React search](https://v7-react.carbondesignsystem.com/?path=/story/components-search--default)

PR #1499, opened 2023-07-31
feat: cv search

davidnixon:
Contributes to #1495

## What did you do?

Port cv-search to Vue3

- [x] Yes

### DIFF
--- a/src/components/CvSearch/CvSearch.stories.mdx
+++ b/src/components/CvSearch/CvSearch.stories.mdx
@@ -4,8 +4,7 @@ import { sbCompPrefix } from '../../global/storybook-utils';
 import { SEARCH_SIZES } from './index';
 import { action } from '@storybook/addon-actions';
 import { ref } from 'vue';
-const myValue = ref('searching');
-
+const myValue = ref('searching galaxies');
 
 <Meta title={`${sbCompPrefix}/CvSearch`} component={CvSearch} />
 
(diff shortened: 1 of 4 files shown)


davidnixon:
@OlkaB could you review this one for me?

OlkaB:
Will do today evening

OlkaB:
@davidnixon adding CvSearch to main.js seems to be missing

OlkaB (commented):
@davidnixon adding CvSearch to main.js seems to be missing

Other than that - all looks good to me 👍

(btw. how detailed reviews you're expecting? Should code work as expected or should I go into details like best code practices / patterns [which has potential to trigger discussions that will extend the time of a reviews] ?)

OlkaB (approved):

davidnixon:
Thank. I really should get rid of main.js. I was using it for some local testing but the real entry point for the build is index.js which uses a regular expression to find all the components during the build.

And yes I'm open to all comments around best practices & patterns. 

FWIW, I do want to turn on more linting for the components. 
currently we use `plugin:vue/vue3-essential` but in other projects I use `plugin:vue/vue3-recommended` which enforces some more best practices. Currently turning that on shows 244 issues across 60 files :-(. I thought maybe best to finish the first full versions of the components before tackling a cleanup round.

OlkaB:
I'm fully supporting the idea of improvements to be implemented after the components are fully ported to Vue3 and all potential issues are being fixed if ineeded. For me it will be a game changer when we will be upgrading large web apps we have in our Company to Vue3 without the need to move them at the same time to new UI lib.

Issue #1450, opened 2023-04-17
CvFileUploader work in progress


felipebritor:
Port CvFileUploader component to vue3
https://vue.carbondesignsystem.com/?path=/story/components-cvfileuploader--default

PR #1509, opened 2023-08-07
feat: cv-file-uploader

felipebritor:
Contributes to #1450

## What did you do?

port cv-file-uploader to vue3.

I decided to leave part of the code a bit 'repeated/duplicated' to make it easier to reason. Is that's not ok, I can change it.

Deeply sorry for the long delay on this one. Had other things coming up these past few months. 

known issues:
- Adjust drag'n'drop area, that is clickable outside the area it supposed to

## How have you tested it?
Mostly, unit tests missing are related to button kind & size and exposed methods

- [ ] N/A
- [ ] No
- [X] Yes

### DIFF
--- a/src/components/CvFileUploader/CvFileUploader.stories.mdx
+++ b/src/components/CvFileUploader/CvFileUploader.stories.mdx
@@ -1,0 +1,333 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs';
+import { sbCompPrefix } from '../../global/storybook-utils';
+import { CvFileUploader, CvFileUploaderSkeleton } from '.';
+import { KINDS } from './const';
+import { action } from '@storybook/addon-actions';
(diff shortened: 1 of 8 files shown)


felipebritor:
Hello, @davidnixon . I'm opening the PR for review, but there is still an style glitch to be solved and a few tests to add. I'll work on it in the next days

felipebritor:
@davidnixon, the last tests are up, the storybook created and the issue mentioned adjusted. 

I'm still unsure if it's the best approach to the issue. In resume: without a constraining style (used at templates on `src/components/CvFileUploader/CvFileUploader.stories.mdx`), the label used for the drop zone is expanding outside the place it's supposed to. It also happens in v2 and a similar solution was implemented.

davidnixon (approved):

Issue #1511, opened 2023-08-09
Work in progress CvStructuredList


OlkaB:
Port CvStructuredList to Vue 3

link to Vue 2 storybook:
https://vue.carbondesignsystem.com/?path=/story/components-cvstructuredlist--default

PR #1513, opened 2023-08-13
feat: port cv-structured-list to Vue3

OlkaB:
Contributes to #1511

## What did you do?
Port CvStructuredList to Vue3

## How have you tested it?
see spec.js files

- [ ] N/A
- [ ] No
- [x] Yes


davidnixon:
Looks really nice. Thanks!

There are some warnings in the console. 
> [Vue warn]: Property "withVModel" was accessed during render but is not defined on instance
> [Vue warn]: Property "listValue" was accessed during render but is not defined on instance.
> runtime-core.esm-bundler.js:41 [Vue warn]: Property "onChange" was accessed during render but is not defined on instance.

Since both stories use the same template the default story still needs to define attributes:
```diff
--- a/src/components/CvStructuredList/CvStructuredList.stories.js
+++ b/src/components/CvStructuredList/CvStructuredList.stories.js
@@ -111,7 +111,10 @@ const Template = args => {
     },
     setup: () => ({
       args,
+      listValue: ref(DefaultListItems[0].id),
       DefaultListItems,
+      onChange: action('change'),
+      withVModel: false,
     }),
     template,
   };
```

davidnixon (changes requested):
see comment

OlkaB:
fixed. thank you @davidnixon

davidnixon (approved):
Thanks!

Issue #1510, opened 2023-08-07
Work in progress cv-time-picker


davidnixon:
**Please read** [Vue 3 guidance here](https://github.com/carbon-design-system/carbon-components-vue/tree/vNext#vue-3-guidance)

Port cv-time-picker to Vue 3

[link to Vue 2 storybook](https://vue.carbondesignsystem.com/?path=/story/components-cvtimepicker--default)

You may want to reference the [Carbon 10 React components](https://v7-react.carbondesignsystem.com/?path=/story/getting-started--welcome)
https://v7-react.carbondesignsystem.com/?path=/story/components-timepicker--default

PR #1518, opened 2023-08-15
feat: port time picker to Vue3

davidnixon:
Contributes to #1510 

## What did you do?

Port `cv-timepicker` including additional `fieldSize` feature found in the React component

Additional work:
- Updated `cv-select` to support features needed in cv-timepicker
- Added `very-strict-eslint.json` which can be used like `npx eslint --config very-strict-eslint.json src/component/Something/CvSomethinhg.vue`. I'll move this to be the default eslint config after all Vue 2 ports are done.
- Remove `main.js` as its not used in the build anymore and causes some confusion

## How have you tested it?

Storybook and jest

- [x] Yes


very-strict-eslint.json
```diff
@@ -0,0 +1,49 @@
+{
+  "root": true,
+  "env": {
+    "node": true
+  },
+  "extends": [
+    "plugin:vue/vue3-recommended",
+    "plugin:prettier-vue/recommended"
```
OlkaB, line 8: is 'eslint:recommended' omitted intentionally?

src/components/CvTimePicker/CvTimePicker.vue
```diff
@@ -0,0 +1,217 @@
...
+      result = props.timezones[0].value;
+      emit('update:timezone', result);
+    }
+  }
+  return result;
+});
+const isDisabled = computed(() => (props.disabled ? true : undefined));
+
+function onInput(e) {
+  emit('change', { type: 'time', value: e?.target?.value });
```
OlkaB, line 206: suggestion for later: changes types might be a candidate to enum (or to be precise js object that tries to pretend to be enum :) ) that will help users to detect change type in a consistent way

eg: 
const TimePickerChangeTypes = {
time: 'time',
ampm: 'ampm',
...
}

OlkaB (approved):

very-strict-eslint.json
davidnixon: No you're right it should be there. I thought it was part of the other 2 but I see [this](https://eslint.vuejs.org/user-guide/#configuration) in the docs.

Issue #1519, opened 2023-08-19
Vue3 File uploader story enhancements


davidnixon:
## Detailed description
Thanks @felipebritor for the port to Vue 3! A couple of asks to improve the storybook a bit. 

- Button story has misleading text
![image](https://github.com/carbon-design-system/carbon-components-vue/assets/7536103/45b87e72-2b96-4422-8e71-e2bdd75c6980)
This is also in the Vue 2 story. Could you change the default text for this story to "Add file" to match whats is in [React](https://v7-react.carbondesignsystem.com/?path=/story/components-fileuploader--file-uploader) 
- The Vue 2 story has this awkward control to show how the state works.
![image](https://github.com/carbon-design-system/carbon-components-vue/assets/7536103/5f9b0272-b661-44c1-bef3-a176f32fe744)
Could you add something better to show how to use the states. The [React](https://v7-react.carbondesignsystem.com/?path=/story/components-fileuploader--file-uploader) story has "Status for file name (filenameStatus)" control so maybe something like that?
- The vue 2 story sets the default accept to ".jpg,.png" which makes it easier to test the invalid case.
![image](https://github.com/carbon-design-system/carbon-components-vue/assets/7536103/dda4204f-3f91-4bad-a2d3-d33b6159f1f4)

## Additional information

The new port to Vue 3 is really nice!

felipebritor:
Hi, @davidnixon . Sorry for skipping these details! I'll work on it along the week

davidnixon:
I think this one is done.

PR #1520, opened 2023-08-20
docs: cv-file-uploader docs enhancements

felipebritor:
Contributes to #1519 

## What did you do?
- ~~Set '.jpg, .png' as initial values for `accept` prop, so users can check invalid state right away~~ (#1529)
- ~~Update button story label to better translate its state~~ (#1529)
- (WIP) allow user to play with the exposed methods

##  Why did you do it?
- Improve user experience / align the current documentation w/ react's

- [X] Yes

### DIFF
--- a/.storybook/manager.js
+++ b/.storybook/manager.js
@@ -7,6 +7,8 @@
 
 import { addons } from '@storybook/addons';
 import theme from './theme';
+import React from 'react';
+import { Form } from '@storybook/components';
 
 addons.setConfig({
   theme,
@@ -16,3 +18,24 @@ addons.setConfig({
 // conditional panels, or other things that get in the way of our workflow
 localStorage.removeItem('@storybook/ui/store');
 localStorage.removeItem('storybook-layout');
+
+/**
+ * This is a **workaround** due to the lack of a 'button' control in storybook.
+ * It relies on postinstall.js defining `window["STORYBOOK_CUSTOM_CONTROLS"]`
+ * as a fallback object, at @storybook/components module, when storybook looks
+ * for a control type.
+ */
+window['STORYBOOK_CUSTOM_CONTROLS'] = {
+  button({ argType, value, onChange }) {
+    return React.createElement(
+      Form.Button,
+      {
+        onClick() {
+          value = typeof value === 'number' ? value : 0;
+          onChange(value + 1);
+        },
+      },
+      argType.buttonLabel || argType.name
+    );
+  },
+};
--- a/.storybook/postinstall.js
+++ b/.storybook/postinstall.js
@@ -1,0 +1,38 @@
+const { resolve } = require('path');
+const { readFileSync, writeFileSync } = require('fs');
+
+/**
+ * This is a fragile **workaround** to customize new storybook controls.
+ *
+ * Currently, storybook lack the option to create customizable controls, as an option for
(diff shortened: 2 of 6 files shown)


felipebritor:
Hi, @davidnixon . Forgive me, again, for the tardiness.

I need your feedback on a few things:

1. I'm persisting the files uploaded in the storybook file as the storybook component is refreshed on some control changes. Unless I'm doing something wrong, as not all control changes reset the component, looks like this is the way to go.
2. `setInvalidMessage` and `setState` are working following the example at react's storybook. Unfortunately, for `clear` and `remove` exposed methods, the native control types doesn't not offer a good option for them. There is [an old issue at storybook repo](https://github.com/storybookjs/storybook/issues/11971) requesting a button control type but not progress. [One of the comments](https://github.com/storybookjs/storybook/issues/11971#issuecomment-1240831460) suggested a workaround which does the trick, but it's quite hacky and fragile. So I need your guidance if it makes sense to use it. We'll have to be mindful of: storybook version must be locked, keep aware that each new version update might break it and update accordingly (even making it look for a `index-hash.js` file does not seems stable enough but might be better than leaving it hardcoded)
You can check it at [f63ad8a](https://github.com/carbon-design-system/carbon-components-vue/pull/1520/commits/f63ad8a71587e1cbd016ebb203415f63ebcee1d0).
If you think it's better not to follow this approach, I can see other options.
image
3. The 'loading' state icon used together with the invalid one is broken (image 2). It's also broken in the v2 and the react version only display a single icon. Should I follow react or adjust the css?
image

felipebritor:
Hi, @davidnixon, just a friendly reminder about the points I need your input ☝️

davidnixon:
>1. I'm persisting the files uploaded in the storybook file as the storybook component is refreshed on some control changes. Unless I'm doing something wrong, as not all control changes reset the component, looks like this is the way to go.

This looks right to me. I know some changes cause the storybook to reload but I'm not sure why. I thought I might take a closer look at it Storybook 7. 

> 2. setInvalidMessage and setState are working following the example at react's storybook. ...

The `clear` and `remove` button in the storybook work correctly for me but I think you are concerned that the control type "function" is [not supported](https://storybook.js.org/docs/react/essentials/controls#annotation)?
```js
  clear: {
    control: 'none',
    type: 'function',
    table: {
      type: { summary: '() => void' },
      category: 'exposed methods',
    },
    description: 'Clear file list',
  },
```

> 3. The 'loading' state icon used together with the invalid one is broken  ...

I think it should be like the react storybook. Only 1 icon is allowed.

felipebritor:
> The clear and remove button in the storybook work correctly for me but I think you are concerned that the control type "function" is [not supported](https://storybook.js.org/docs/react/essentials/controls#annotation)?

`function` would be the type and `button` the control type. [This control type does not exist and is set as feature request for some time now](https://github.com/storybookjs/storybook/issues/11971). What I did was setup a "hacky" custom button type at [1e28477](https://github.com/carbon-design-system/carbon-components-vue/pull/1520/commits/1e28477c2d5ea5c70333c8449eaacc492ff0a254).  

> I think it should be like the react storybook. Only 1 icon is allowed.

Done at [ff4241f](https://github.com/carbon-design-system/carbon-components-vue/pull/1520/commits/ff4241f961cff102f9943f70232e442c8ecad7b1) 👍

davidnixon (approved):

Issue #1522, opened 2023-08-30
Add slot to gain control over tab's switching buttons


OlkaB:
**Is your feature request related to a problem? Please describe.**
I needed to create a separate component (basically same as CvTabs) only to add slots to have more control over tab buttons content.

**Describe the solution you'd like**
add slot for customizing tab button content

OlkaB:
working on it

PR #1523, opened 2023-08-30
feat: add slot to gain control over tabs switching buttons

OlkaB:
Contributes to #1522

## What did you do?
Added slot to customise tab switching buttons' content

## How have you tested it?
added test for slot in CvTabs.spec.js

- [x] Yes

sample:
image


davidnixon:
I like this solution but I think using the label is a problem. The Label would often be localized so the "House" label might become "Taigh" or "Maison" or "Casa". Could you use the id instead?

OlkaB:
fair point - change applied

davidnixon (commented):
sooooooo close:
"To customise tab look, in CvTabs component use scoped slot named with tab's **label**. Example: current tab has **label** "**House**", so the code for customising tab looks like:"

need to switch text to **id** :smile:

OlkaB:
oh gosh - truly. Done, thanks @davidnixon

davidnixon (approved):
Thanks!

Issue #1524, opened 2023-09-07
Vue3: CvOverflowMenu position is calculated incorrectly when used inside CvDataTable cell


OlkaB:
## Description
When using CvOverflowMenu within CvDataTableCell, the menu's dropdown position is calculated far outside the table
It should behave like in Vue2 (e.g. see this story https://vue.carbondesignsystem.com/?path=/story/components-cvdatatable--slotted-html)
Browser: Chrome Version 116.0.5845.180
Carbon Design System version: Vue3

## Steps to reproduce the issue
import CvOverflowMenu and CvOverflowMenuItem to CvDataTable.stories.mdx and replace content one of the cells like so:
```
<cv-data-table-cell>
          <cv-overflow-menu flip-menu style="margin: 0 auto;">
            <cv-overflow-menu-item>Edit</cv-overflow-menu-item>
            <cv-overflow-menu-item>Test</cv-overflow-menu-item>
          </cv-overflow-menu>
      </cv-data-table-cell>
```
![image](https://github.com/carbon-design-system/carbon-components-vue/assets/20342514/fd7ce92c-834c-4b15-b894-3e7e79bc38db)

## Additional information
at start table looks correctly:
![image](https://github.com/carbon-design-system/carbon-components-vue/assets/20342514/a175efde-24db-440e-ab89-cf922abd92d6)

but after clicking three dots menu button this happens:
![image](https://github.com/carbon-design-system/carbon-components-vue/assets/20342514/0b1192d4-ec06-429e-b109-795a39ca8add)

PR #1525, opened 2023-09-07
fix: overflow menu dropdown placement and position

OlkaB:
Contributes to #1524

## What did you do?
Fixed placement and position for cv-overflow-menu so it will behave as in Vue2 (the menu's dropdown is appended to document.body)

- [x] Yes


OlkaB:
@davidnixon If the new placement of the menu was intentional - please discard this PR. In such case, the issue still would need a different fix (potentially it might be the one mentioned in the ticket #1524 comment, but haven't tested it widely)

davidnixon (commented):
lgtm

davidnixon (approved):
lgtm

Issue #1526, opened 2023-09-08
Work in progress [cv-tag-skeleton]


caioafc:
**Please read** [Vue 3 guidance here](https://github.com/carbon-design-system/carbon-components-vue/tree/vNext#vue-3-guidance)

Port [cv-some-component] to Vue 3

Add link to Vue 2 storybook here. For example:
https://vue.carbondesignsystem.com/?path=/story/components-cvtoggle--default

You may want to reference the [Carbon 10 React components](https://v7-react.carbondesignsystem.com/?path=/story/getting-started--welcome)

PR #1528, opened 2023-09-13
feat: CvTag / CvTagSkeleton changes

caioafc:
Contributes to https://github.com/carbon-design-system/carbon-components-vue/issues/1526

### CvTag changes
-  Added `small` property to change component size (chosen `small` instead `size` because is the only available option for this component)

- Created end tag for skeleton div, `<div/>` is not the best option

- Reorganized initial tagClasses, we doesn't need to set `kind` classes when skeleton component is being used
- Set label on Skeleton scenario in the Storybook (fixing console warn above)
  image

### CvTagSkeleton
- Created CvTagSkeleton for migration purposes by having a separated component to use Skeleton (preserving Vue 2 pattern)

## How have you tested it?
- Created unit tests
- Checked Storybook behaviors and it's console

- [ ] N/A
- [ ] No
- [x] Yes

### DIFF
--- a/src/components/CvTag/CvTag.vue
+++ b/src/components/CvTag/CvTag.vue
@@ -56,6 +56,10 @@ export default {
      * skeleton used when loading
      */
     skeleton: Boolean,
+    /**
+     * tag size small
+     */
+    small: Boolean,
   },
   emits: [
     /**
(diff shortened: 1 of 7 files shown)


src/components/CvTag/CvTag.vue
```diff
@@ -56,6 +56,10 @@ export default {
      * skeleton used when loading
      */
     skeleton: Boolean,
+    /**
+     * tag size small
+     */
+    small: Boolean,
```
davidnixon, line 62: :+1:

davidnixon (approved):
lgtm. thanks!

Issue #1538, opened 2023-09-28
Add accessibility test


davidnixon:
Many components need a basic accessibility test. See [guidance](https://github.com/carbon-design-system/carbon-components-vue/blob/main/ACCESSIBILITY-TESTING.md)

Add accessibility test  (not all at once). Either create a work item or reference this issue.

Let's start with these components for Accessibility tests:

- [ ] CvHeaderGlobalAction
- [ ] CvFileUploader
- [ ] CvButtonSet
- [ ] CvMultiSelect
- [ ] CvLink
- [ ] CvList
- [ ] CvCheckbox
- [ ] CvTextArea
- [ ] CvTextInput
- [ ] CvRadioButton
- [ ] CvRadioGroup

halivert:
May I take `CvLink` one?

PR #1539, opened 2023-10-04
test(cv-link): add tests for a11y

halivert:
Contributes to #1538 

## What did you do?
Add a11y test for `CvLink`

## Why did you do it?
Adding 3 test cases, for simple link, one in disabled state and the other one with an icon

## How have you tested it?
Using testing library, the icon one is failing because needs a label, but I'm not sure how to proceed

### Notes
I have an idea of naming convention for accessibility tests, maybe we can use the abbreviation, `[Component].a11y.spec.js`?

Thank you in advance

### DIFF
--- a/src/components/CvLink/CvLink.vue
+++ b/src/components/CvLink/CvLink.vue
@@ -14,7 +14,12 @@
     ]"
   >
     <slot></slot>
-    <CvSvg v-if="icon" :class="`${carbonPrefix}--link__icon`" :svg="icon" />
+    <CvSvg
+      v-if="icon"
+      :class="`${carbonPrefix}--link__icon`"
+      :svg="icon"
+      alt=""
+    />
   </component>
 </template>
 
--- a/src/components/CvLink/__tests__/CvLink.accessibility.spec.js
+++ b/src/components/CvLink/__tests__/CvLink.accessibility.spec.js
@@ -1,0 +1,34 @@
+import { CvLink } from '..';
+import { render } from '@testing-library/vue';
+
+describe('CvLink - accessibility', () => {
+  it('CvLink - basic', async () => {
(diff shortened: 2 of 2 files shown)


davidnixon:
I updated the guidance a bit to add a link to the React storybook. There might be DOM difference for accessibility there that might be helpful in resolving any test issues.
> The accessability rules seems to have changed since Carbon 10 was relesed but still it might be helpful to reference the [React Carbon 10 Storybook](https://v7-react.carbondesignsystem.com/?path=/story/getting-started--welcome) to see if there are difference in the DOM for accessability.

davidnixon:
I think the test cases can be resolved by adding `alt=""`  to the component
https://github.com/carbon-design-system/carbon-components-vue/blob/337cb41677bb308c20c43938f3485faaf7db87c0/src/components/CvLink/CvLink.vue#L17

halivert:
Thanks @davidnixon, I wasn't sure if I should update actual component, and I had a lot of work lately, but thanks for your advice 👍🏽 
Done in 0d55fcc7c999e9dd1ba418b4302330b1438cab5a

src/components/CvLink/CvLink.vue
```diff
@@ -14,7 +14,12 @@
     ]"
   >
     <slot></slot>
-    <CvSvg v-if="icon" :class="`${carbonPrefix}--link__icon`" :svg="icon" />
+    <CvSvg
+      v-if="icon"
+      :class="`${carbonPrefix}--link__icon`"
+      :svg="icon"
+      alt=""
```
halivert, line 21: Empty alt because it's "an advisory icon of minimal importance"

davidnixon (approved):
lgtm

PR #1545, opened 2023-10-18
chore: migration lint

davidnixon:
## What did you do?
- enable very strict eslint linting:
   ```json
    "extends": [
      "eslint:recommended",
      "plugin:vue/vue3-recommended",
      "plugin:prettier-vue/recommended"
    ],
   ```
- automatically fix eslint issue with eslint **--fix** src/**/*.{js,vue}
- fix remaining lint issues. Mostly this is about assigning default values to properties:
   ```js
    sampleString: { type: String, default: undefined },
    sampleArray: { type: Array, default: () => [] }, // mostly but sometimes undefined
    sampleObject: { type: Object, default: undefined } // mostly but sometimes Object()
   ```
- suppress warning in test like `[Vue warn]: injection "selectable" not found.`
- When using Vue compatibility mode to migrate a Vue 2 app to Vue 3 with Vue 3 Carbon Components. Some warnings are shown in the console like `[Vue warn]: (deprecation ATTR_FALSE_VALUE) Attribute "aria-invalid" ...`. To resolve many templates are updated like this:
   ```diff
   -        :data-invalid="data.isInvalid"
   +        :data-invalid="data.isInvalid || null"
   ```
   See [BREAKING: No longer removes attribute if ...](https://v3-migration.vuejs.org/breaking-changes/attribute-coercion.html)

## Why did you do it?
- very strict linting had been on hold until after component migration was complete.
- working on migration Vue 2 apps now through the end of the year so there may be more warning uncovered by the compatibility compiler. 

## How have you tested it?

With local app and the jest tests

- [x] N/A


davidnixon:
@OlkaB could you review this one? It's a lot of changes but most are related to the strict linting. 
Could you run this branch locally and just make sure it looks OK?

OlkaB:
@davidnixon With pleasure, but could you please let me know how may I checkout your branch?

mateusbandeiraa:
I also noticed that the linting during the commit hook is trying to fix this file: https://github.com/carbon-design-system/carbon-components-vue/blob/main/ACCESSIBILITY-TESTING.md

Would this PR be a good candidate to fix the file?

src/components/CvSearch/CvSearch.vue
```diff
@@ -33,20 +33,20 @@
...
+        ref="elInput"
         v-model="internalSearchText"
-        @input="onInput"
+        role="searchbox"
+        :class="`${carbonPrefix}--search-input`"
         type="text"
-        ref="elInput"
         :placeholder="placeholder"
         :aria-labelledby="uid"
+        :disabled="disabled ? 'true' : undefined"
```
OlkaB, line 45: following compatibility mode comment, might be that also here it should be changed to :disabled="disabled || null" ?

src/components/CvTextArea/CvTextArea.vue
```diff
@@ -14,16 +14,16 @@
...
       <WarningFilled16
         v-if="isInvalid"
         :class="`${carbonPrefix}--text-area__invalid-icon`"
       />
       <textarea
-        :aria-invalid="isInvalid"
-        :aria-describedby="isInvalid ? errorId : undefined"
         :id="cvId"
+        :aria-invalid="isInvalid || null"
+        :aria-describedby="isInvalid ? errorId : undefined"
```
OlkaB, line 26: shouldn't there be null instead of undefined?

src/components/CvButton/CvIconButton.vue
```diff
@@ -30,8 +30,7 @@ const { disabled, icon, kind, size } = commonCvButtonProps;
```
OlkaB, line 30: I'm getting this warning:
image

The value received is an empty string, but alsp seems this prop/attribute is not used in the story at all

src/components/CvGrid/CvColumn.vue
```diff
@@ -13,22 +13,27 @@ const props = defineProps({
   sm: {
```
OlkaB, line 13: side note for this component story: it's not working
image

src/components/CvPagination/CvPagination.vue
```diff
@@ -2,12 +2,12 @@
   <div :class="`cv-pagination ${carbonPrefix}--pagination`" data-pagination>
```
OlkaB, line 2: side note: pagination styling seem to fail
image

OlkaB:
@davidnixon Review finished. I've caught some additional issues but they might be better candidates to separate issue tickets.

src/components/CvButton/CvIconButton.vue
davidnixon: Updated story
1. Get tip alignments fom '../CvTooltip/consts.js';
2. Fix warnings from story book about deprecated format.
3. Add control for tipAlignment

src/components/CvGrid/CvColumn.vue
davidnixon: This one is **definitely** related to the lint changes. For the lint errors I added  `default: undefined` for the break-point size and this is causing the error. Updated to  `default: false`

src/components/CvPagination/CvPagination.vue
davidnixon: This is unrelated to this PR but I "fixed" it. 
@felipebritor looks like you added this to `.storybook/styles.scss` 
```css
  // set space between form items
  .bx--form-item {
    margin-bottom: 2rem;
  }
```
I'm sure I broke something by removing it so maybe have a look and add it to a story instead of to the style?

Issue #591, opened 2019-10-11
Export consts for properties like kind, state etc from each component.


lee-chase:
A number of components have fixed values for their properties such as the kind of a CvButton. It would be nice if these were exported to consumers.

PR #1561, opened 2023-12-08
fix: fixes related to migration/compat build

davidnixon:
## What did you do?

- avoid more warning related to attributes using boolean values. Attributes that need "true" or "false" string now use strings. Attributes that are supposed to be removed if the boolean is false now use "boolean || null"
- add constants for modal kinds and size. This fits  with strategy in #591 
- issue console warning for validation failure in modal
- watching an array require "deep" to be set to true
- The inline notification subtitle supports html in Vue 2 but not Vue 3. Add a slot for user (like me) who need this feature. Also added slot for `title`. Also add doc for the slots to the storybook.

## Why did you do it?

Compiling another app with vue migration/compatibility mode revealed these issues

- [x] Yes

### DIFF
--- a/src/components/CvDatePicker/CvDatePicker.vue
+++ b/src/components/CvDatePicker/CvDatePicker.vue
@@ -48,7 +48,6 @@
             :pattern="pattern"
             :placeholder="placeholder"
             @change="handleUpdateEvent"
-            @click="handleClick"
           />
           <Calendar16
             v-if="['single', 'range'].includes(getKind)"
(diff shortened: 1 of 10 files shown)


davidnixon:
@Thenafi can you tell me what you think of this change? Just need someone to review it before merging.

davidnixon:
@ OlkaB
Could you look at this one and lmk if its Ok?

OlkaB (approved):
Looks good to me

Issue #1560, opened 2023-11-27
Sorting Issue with CvDataTable in Carbon Design System + Vue 3


hgwrtsdev:
Hi Carbon Design System team,

**Description:**
I am urgently reaching out regarding a critical issue with the CvDataTable component in the Carbon Design System + Vue 3. The sorting functionality, expected to be triggered when clicking on the column headers, is malfunctioning. Despite the sorting icons transitioning, the data remains unsorted.

**Steps to Reproduce:**

1. Visit the official documentation/demo for Carbon Design System + Vue 3: https://vue.carbondesignsystem.com/.
2. Navigate to the CvDataTable section.
3. Click on the column headers to trigger sorting.

**Expected Behavior:**
The data in the CvDataTable should be sorted based on the selected column.

**Actual Behavior:**
The sorting icons transition, but the data does not get sorted.

**Environment:**
Carbon Components version: ^10.58.10
@carbon/vue version: ^3.0.9
Vue version: ^3.2.13

**Additional Context:**
This sorting functionality is crucial for our project, and we rely heavily on the CvDataTable component. Unfortunately, downgrading to version 2+ is not a viable solution due to compatibility issues with other dependencies.

I appreciate your attention to this matter, and I understand the importance of ensuring a smooth experience with Carbon Design System components. Please let me know if any additional information is needed or if I can assist in any way.
Or if anyone has the same problem, how did you solve it?

davidnixon:
done.
https://vue.carbondesignsystem.com/?path=/story/component-cvdatatable--default-story

PR #1563, opened 2023-12-08
fix: sort in datatable story.

davidnixon:
Contributes to #1560 

## What did you do?

Added a sort function to the story and more documentation tot he story about the need for user's to implement their own search.

## Why did you do it?

The current story emits the search events but does not actually sort the data. This is confusing.

- [x] Yes


davidnixon:
Initial state
> image


After clicking on "name" header
> image


After clicking on "year" heading
> image

After clicking on "year" heading again (descending sort)
> image

After clicking on "year" heading a 3rd time (no sort)
> image

davidnixon:
@felipebritor what do you think of this change?

felipebritor:
`.sort` mutate the sorted array, which is global, but doesn't look like an issue.  LGTM, @davidnixon 👍

PR #1568, opened 2024-01-28
feat: accordion v-model

davidnixon:
## What did you do?

Added a v-model for accordion items open state

## Why did you do it?

I have a design that calls for an "auto close" of sections such that when one section is opened, other sections close. There is not a way to implement this in current component.

## How have you tested it?

With local story and in test app that implements the "auto close"

- [x] Yes


davidnixon:
FWIW, auto close code is included in the "docs" arae of the storybook
```js
// example auto close
// <cv-accordion @change="onAutoClose">
const open = ref({
  accItem1: false,
  accItem2: false,
  accItem3: false,
  accItem4: false,
});
function autoClose(ev) {
  if (ev.change.open) {
    for (let state in open.value) {
      if (state !== ev.change.id) open.value[state] = false;
    }
  }
}
```

davidnixon:
@OlkaB could you review this one?

src/components/CvAccordion/CvAccordion.vue
```diff
@@ -83,6 +83,19 @@ provide('registerItem', (itemCvId, item) => {
 provide('deregisterItem', itemCvId => {
   items.delete(itemCvId);
 });
+
+/**
+ * @typedef {Object} AccordianChangeElement
```
OlkaB, line 88: side note: typo in Accordion word in whole types section
is: Accordi**a**n     =>     should be: Accordi**o**n

src/components/CvAccordion/consts.js
```diff
@@ -1,16 +1,16 @@
 export const alignConsts = [null, 'start', 'end', ''];
 alignConsts.$labels = {
-  'Not supplied': 0,
-  Start: 1,
-  End: 2,
-  'Empty string': 3,
+  'Not supplied (end)': 0,
```
OlkaB, line 3: I wonder if it would be more understandable to replace both "empty string" and "null" options with just single "default" or to be more precise "default (start)"? 
For me the meaning of (especially) "not supplied" was not fully clear

OlkaB (approved):
Left 2 side comments, but other than that looks good to me👍

src/components/CvAccordion/consts.js
davidnixon: image

image

davidnixon: Yes that was confusing. Love the suggestion.

src/components/CvAccordion/CvAccordion.vue
davidnixon: Nice find! Thanks!

OlkaB (approved):

PR #1570, opened 2024-02-03
fix: a11y tests for CvList, CvCheckbox, CvRadioButton, CvRadioGroup

benceszenassy:
Contributes to #1538 

## What did you do?

Add CvList, CvCheckbox, CvRadioButton, CvRadioGroup accessibility tests and fix components and tests if needed.

## Why did you do it?

To make component more accessible.

## How have you tested it?

With the `expect(...).toBeAccessible(...)` built in utility.

- [ ] N/A
- [ ] No
- [x] Yes

## CvRadioGroup

This component did not have a `fieldset`, `legend` pair, my solution was based on the [v10 react component](https://github.com/carbon-design-system/carbon/blob/v10/packages/react/src/components/RadioButtonGroup/RadioButtonGroup.js#L177).

The tamplate changed from this:
image
To this:
image

### DIFF
--- a/src/components/CvCheckbox/CvCheckbox.vue
+++ b/src/components/CvCheckbox/CvCheckbox.vue
@@ -13,7 +13,6 @@
       type="checkbox"
       :class="`${carbonPrefix}--checkbox`"
       :checked="isChecked === true"
-      :aria-checked="`${isChecked}`"
       :value="value"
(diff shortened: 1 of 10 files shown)


davidnixon (approved):
very nice! lgtm

PR #1571, opened 2024-02-12
fix: date picker input backspace

davidnixon:
Contributes to #1574

## What did you do?
Resolve question from slack
> Developer tools in Google Chrome has an error when deleting the value of the CvDatePicker(kind = single) in the Backspace Key. (I'm using Windows, I haven't tried Mac)
In this case, property bind v-model of CvDatePicker it will not be cleared for property, but appearance will be cleared.
And the I think CvDatePicker Stroybook having the same problem. (similar errors will be displayed in the developer tools)
Storybook : https://vue.carbondesignsystem.com/?path=/story/component-cvdatepicker--single-using-date

## How have you tested it?

Add a v-model to the "single-using-date" story to facilitate testing 

- [x] Yes

### DIFF
--- a/src/components/CvDatePicker/CvDatePicker.stories.js
+++ b/src/components/CvDatePicker/CvDatePicker.stories.js
@@ -31,17 +31,11 @@ export default {
     },
   },
   argTypes: {
-    dateLabel: { type: String },
-    dateEndLabel: { type: String },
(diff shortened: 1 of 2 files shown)


davidnixon:
I see from slack that with the provided fix the issue:
> has been resolved in my local environment.

Issue #1578, opened 2024-03-14
Vue3 - missing (?) named components' exports from the library


OlkaB:
In Vue2 there was an option to import components separately from the library (instead of installing everything) by using a named imports
``` 
import { CvButton } from '@carbon/vue';
```

This way doesn't work in Vue3 anymore. Seems there is only default export of install components plugin.
**Question**: Is this intentional? Is there a change to restore also named exports?

davidnixon:
I have not tried that and I didn't know that was an option in Vue2.
I did know that you could selectively  import items like:
```js
app.use(CarbonVue3, ["CvButton", etc]);
```
But I have not tried that either. That code is here:
https://github.com/carbon-design-system/carbon-components-vue/blob/066aac04cd806f22bdb8458032f5b96afb09f5a3/src/index.js#L27-L32

I see this in Vue 2 branch
https://github.com/carbon-design-system/carbon-components-vue/blob/c63f589bfc983ab1b46afa8736c5857e68f96cdc/packages/core/src/index.js#L28-L30
...

I'm sure we could replicate that with a script so that it would be a little more maintainable.

PR #1580, opened 2024-03-19
chore: create script to add named exports of a components to index.js

OlkaB:
Contributes to #1578

## What did you do?
Added prebuild script to auto add named exports of components

## Why did you do it?
This was option available in Vue2 and change from it large repo requires additional piece of work, so prefer to restore it at source

## How have you tested it?
run build script then generated local tgz package and tested in Vue3 repo

- [X] N/A


davidnixon (changes requested):
@OlkaB Sorry I lost track of this one.
First, this is awesome work. Thank you!

1. `convertToImportPathRelativeToIndex` does not work for me. It always returns null which means no exports are added.  I do not understand the regex used here so not sure what exactly is wrong but I did confim that this code works ok:
```js
   const indexDir = path.dirname(INDEX_FILE_PATH);
   return './' + path.relative(indexDir, systemPath);
```

With that I can do something like 
```js
<wow>I'm a button</wow>
...
import { CvButton as Wow } from "@carbon/vue";
```

2. I think only components that start with "Cv" should be included.  The others are not intended for direct import. For example, these shuold be excluded:
- `./components/CvDataTable/_CvDataTableRowInner.vue`
- `./components/CvSvg/__tests__/_TestIcon.vue`
- `./components/CvGrid/_SbContainer.vue`
3. The "prebuild" hook is not run automatically for me. Can you add to "ci-check" in pcakage.json?

OlkaB:
@davidnixon thank you for a review. 

1. I'm not sure where should I place the code you've provided
```
   const indexDir = path.dirname(INDEX_FILE_PATH);
   return './' + path.relative(indexDir, systemPath);
```
  the only place, which provides `systemPath` variable is `convertToImportPathRelativeToIndex` method. Is this a place which you meant (so match should use the code returned by what you wrote)?
  Sorry for asking, but I do not know how GH pipelines works or how to check it, as locally it works for me (which seems not to be helpful in this case😌).
  
  2. addressed - underscored components are now not included
  3. I've moved script to ci-check as you've suggested
  
  btw. this is what I receive in ./src/index.js when running locally 
  
image

davidnixon:
Sorry yeah I just replaced the contents of convertToImportPathRelativeToIndex with
```js
function convertToImportPathRelativeToIndex(systemPath) {
   const indexDir = path.dirname(INDEX_FILE_PATH);
   return './' + path.relative(indexDir, systemPath);
}
```
That works for me locally. I think I see the issue. You are on a windows box I think? So this regex
`const UNNECESSARY_SYSTEM_PATH_CHUNK_REGEX = /.*\\src\\(.*)/;`
works if I flip to \\ to \/. If the `path.relative` I think that's a clear look to the function.

OlkaB:
Indeed - I'm working on Windows machine. Thank you very much for the help and explanation. I've made a changes as suggested.

davidnixon (approved):
Works on Mac and Linux. 
Tested in seperate app like:
```js
<template>
        <wow>I'm a button</wow>
        <awesome>I'm a toggle</awesome>
</template>
<script setup>
import { CvButton as Wow, CvToggle as Awesome } from "@carbon/vue";
...
```
image

LGTM.

Issue #1581, opened 2024-03-22
Some components in Vue3 are missing '.cv-{componentName}' class


OlkaB:
## Detailed description
Some components in Vue3 are missing '.cv-{componentName}' class

## Steps to reproduce the issue
Example:
Vue3 docs (inspect element) [https://vue.carbondesignsystem.com/?path=/story/component-cvbuttons-cvbutton--field]
vs Vue2 docs [https://vue.carbondesignsystem.com/vue2/?path=/story/components-cvbutton--default]

## Additional information

Vue2 
![image](https://github.com/carbon-design-system/carbon-components-vue/assets/20342514/be10fe9a-bf89-480a-ba26-9e18875f8da9)

Vue3
![image](https://github.com/carbon-design-system/carbon-components-vue/assets/20342514/4de7efd6-553e-4144-949f-f1ef3d3d9293)

OlkaB:
completed

PR #1584, opened 2024-04-05
fix: replicate missing `cv-` classes from Vue2 repo

OlkaB:
Contributes to #1581

## What did you do?
Replicated missing `cv-` classes that were present on the components in Vue2

## Why did you do it?
To minimise amount of changes we need to do when migrating to Vue3

## How have you tested it?
run updated tests

- [x] N/A
- [ ] No
- [ ] Yes


davidnixon (approved):
lgtm

Issue #1588, opened 2024-04-15
CvTag in Vue3 has default color 'red' while in Vue2 it was one of grays


OlkaB:
## Detailed description

CvTag in Vue3 has default color 'red' while in Vue2 it was one of grays, which complicates the migration

## Steps to reproduce the issue

please compare storybook's https://vue.carbondesignsystem.com/vue2/?path=/story/components-cvtag--standard vs https://vue.carbondesignsystem.com/?path=/docs/component-cvtag--default

PR #1589, opened 2024-04-15
fix: restore CvTag default color from Vue2

OlkaB:
Contributes to #1588

## What did you do?
Restored CvTag default color used in Vue2 

## Why did you do it?
To preserve consistency for when migrating from Vue2 to Vue3. Also red is not the color that is the best as default value (is not neutral)

## How have you tested it?
Run unit tests; visually on served Storybook
image

- [x] Yes


OlkaB:
@davidnixon After this PR is approved & merged could I ask for releasing latest changes to production?

davidnixon (approved):

PR #1592, opened 2024-04-19
Fix publish

davidnixon:


davidnixon:
@OlkaB I ended up having to commit the index.js with the export. #1578 
So I think we need a way now to remove them so they will re-generate. I think I have something for this but any ideas are welcome.

OlkaB:
Hi @davidnixon, thanks for the information. Let me please clarify: do I understand correctly we need a separate script to remove the added components' exports?

PR #1593, opened 2024-04-26
fix: ibmtelemetry not run in prod

davidnixon:
## What did you do?

The `ibmtelemetry` should be part of postinstall but it was not. This PR, hopefully, fixes that.

## How have you tested it?
```sh
yarn build
yarn pack
tar -xvf package.tgz
jq .scripts.postinstall package/package.json
```
extract `package/package.json` from generated package.tgz
`"postinstall": "ibmtelemetry --config=telemetry.yml",`

Also verified that package.json is unchanged after running these commands.

The root cause of this issue is [pinst](https://www.npmjs.com/package/pinst) npm package.
Originally the `postinstall` script was meant only for developers of this repo. It installed some git pre-commit hooks. Obviously that should not happen for users of this package and `pinst` did a nice job of preventing the `postinstall` from running for package users. 

Now, however, we need two different `postinstall` scripts.
1. postinstall that runs git pre-commit hooks - this is dev only
2. postinstall that runs production hooks like ibmtelemetry
The `pinst` package does not handle this use case. So I added the `packrat.js` that toggles between these 2 versions of `postinstall`.

Another possible solution is to stop using yarn. :-( The npm package manager has a `prepare` [lifecycle hook](https://docs.npmjs.com/cli/v10/using-npm/scripts) that could accommodate this use case. Also, pnpm has a very nice [devPreinstall hook](https://pnpm.io/scripts) which would work.

Please ignore the noisy .yarn/cache updates


package.json
```diff
@@ -23,6 +24,7 @@
     "lint:es": "eslint src/**/*.{js,vue}",
     "lint:style": "vue-cli-service lint:style",
     "serve:storybook": "cd ./storybook && yarn run serve",
+    "prepare": "echo hello lifecycle hook",
```
francinelucca, line 27: just making sure this change is intended here

francinelucca (approved):
LGTM! thanksss

package.json
davidnixon: I'll remove that. Just testing.

Issue #1594, opened 2024-04-29
CvCheckbox is not loosing focus style when clicked outside [Vue3]


OlkaB:
## Detailed description
CvCheckbox is not loosing focus style (blue outline or more precisely 'bx--checkbox-label__focus' class) when clicked outside.
In Vue2 it happened.

## Steps to reproduce the issue

1. go to https://vue.carbondesignsystem.com/?path=/story/component-cvcheckbox--default
2. Select checkbox
3. Click somewhere outside
4. See the blue outline hasn't disapeared
5. Compare with Vue2 https://vue.carbondesignsystem.com/vue2/?path=/story/components-cvcheckbox--default

## Additional information
![image](https://github.com/carbon-design-system/carbon-components-vue/assets/20342514/1789ff5c-b02e-4619-baac-4730530f8fc2)

PR #1595, opened 2024-04-29
fix: unfocus CvCheckbox (add missing 'value' access to ref)

OlkaB:
Contributes to #1594

## What did you do?
Fix: unfocus checkbox style on blur

## Why did you do it?
Te code was broken

## How have you tested it?
Visually, run unit tests for CvCheckbox

- [x] N/A


benceszenassy (approved):
Nice catch!

davidnixon (approved):

Issue #1573, opened 2024-02-17
<cv-file-uploader> Failed to resolve import


krko12345:
Failed to resolve import

## Detailed description

> Describe in detail the issue you're having.

CvFileUploader throws error  when used in my project. The error is "Failed to resolve import", which is related to CvForm component import in the file-uploader component.

In my env.d.ts  I have declared the path as "declare module '@carbon/vue/src/components/*'", and I tried with different variations as well.

Still getting error:
![image](https://github.com/carbon-design-system/carbon-components-vue/assets/36237241/58ef5168-64f3-4fd6-b307-367899c5142f)

What works for me is to modify the CvFileUploader component and remove the CvForm wrapper inside, but of course that's not permanent solution.

Any recommendation? Is this a bug?

> Is this a feature request (new component, new icon), a bug, or a general issue?
issue

> Is this issue related to a specific component?
CvFileUploader

> What did you expect to happen? What happened instead? What would you like to see changed?
Expected the component to load, instead it throws error.

> What browser are you working in?
Edge

> What version of the Carbon Design System are you using?
3.0.12

> What offering/product do you work on? Any pressing ship or release dates we should be aware of?

## Steps to reproduce the issue

1. Step one
2. Step two
3. Step three
4. etc.

## Additional information

- Screenshots or code
- Notes

krko12345:
The problem seems to move to next line if I modify it to use parent directory path (instead of current directory) path for the import

![image](https://github.com/carbon-design-system/carbon-components-vue/assets/36237241/d5ed20e6-0c5c-4751-b607-0660c92fe453)

davidnixon:
Hmm I cannot reproduce this. I think you are correct that its may have something to do with 'env.d.ts' file. What happens if you remove the "declare module " statement?

krko12345:
Error remained even after removal of all declarations from "env.d.ts". 
It wouldn't be a solution anyway because without declarations the code would fail to compile from typescript.

krko12345:
Not sure what more should I include in my response. The problem still persists as described.

I only managed to avoid the problem by having a modified copy of CvFileUploader component, where I have removed the CvForm wrapper. As mentioned, other carbon components work for me as expected.

davidnixon:
I do not think I fully understand the use case but I think I do see the issue. Several index files import like this:
```js
import CvForm from './CvForm';
```
Notice the lack of ".vue" extension! 
I updated those places to import like this:
```js
[...]

PR #1606, opened 2024-06-18
fix: import with vue extension

davidnixon:
Contributes to #1573 

## What did you do?
add vue extension to import statements where it was missing.

## Why did you do it?
This **may** resolve 1573

## How have you tested it?
Cannot reproduce but sent to issue author for testing.

- [x] N/A


davidnixon:
See review in #1573

PR #1610, opened 2024-07-09
fix: created index.js for cv-grid and related components

### LINKED ISSUE #1608: Component CvGrid is missing the index.js
Component CvGrid is missing the index.js so the normal import of CvGrid, CvColumn and CvRow doesn't work installing the package

1. Install @carbon/vue
2. Try to import @carbon/vue/src/components/CvGrid

oskarnrk:
This PR fixes #1608.
I just added the CvGrid's index.js that was missing. It was not possibile to import and use CvGrid and its components otherwise.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@carbon/vue",
   "license": "Apache-2.0",
-  "version": "3.0.17-wp01",
+  "version": "3.0.17",
   "description": "A collection of components for the Carbon Design System built using Vue.js",
   "packageManager": "yarn@3.2.0",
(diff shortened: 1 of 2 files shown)


davidnixon:
I think this is related to #1578 
Your use case is 
```js
import { CvGrid } from "@carbon/vue";
```
correct?

davidnixon (approved):

oskarnrk:
> I think this is related to #1578 Your use case is
> 
> ```js
> import { CvGrid } from "@carbon/vue";
> ```
> 
> correct?

Correct. Thanks for your approval and merge!

Issue #1576, opened 2024-02-26
Cannot install carbon-vue on fresh Vue 3 project


TNSaturday:
## Detailed description

> Describe in detail the issue you're having.

I'm trying to install @carbon/vue 3 on a new Vue 3 project generated using Vite Vue 3 template.

> Is this a feature request (new component, new icon), a bug, or a general issue?

Bug.

> Is this issue related to a specific component?

No.

> What did you expect to happen? What happened instead? What would you like to see changed?

I expect `npm add` command to run succesfully. Instead got this error message:

```
npm WARN cleanup Failed to remove some directories [
npm WARN cleanup   [
npm WARN cleanup     'C:\\dev\\vue-carbon-emit\\node_modules\\carbon-components',
npm WARN cleanup     [Error: EBUSY: resource busy or locked, rmdir 'C:\dev\vue-carbon-emit\node_modules\carbon-components'] {
npm WARN cleanup       errno: -4082,
npm WARN cleanup       code: 'EBUSY',
npm WARN cleanup       syscall: 'rmdir',
npm WARN cleanup       path: 'C:\\dev\\vue-carbon-emit\\node_modules\\carbon-components'
npm WARN cleanup     }
npm WARN cleanup   ],
npm WARN cleanup   [
npm WARN cleanup     'C:\\dev\\vue-carbon-emit\\node_modules\\@vueuse\\core',
npm WARN cleanup     [Error: EBUSY: resource busy or locked, rmdir 'C:\dev\vue-carbon-emit\node_modules\@vueuse\core\node_modules\vue-demi'] {
npm WARN cleanup       errno: -4082,
npm WARN cleanup       code: 'EBUSY',
npm WARN cleanup       syscall: 'rmdir',
npm WARN cleanup       path: 'C:\\dev\\vue-carbon-emit\\node_modules\\@vueuse\\core\\node_modules\\vue-demi'
npm WARN cleanup     }
npm WARN cleanup   ],
npm WARN cleanup   [
npm WARN cleanup     'C:\\dev\\vue-carbon-emit\\node_modules\\@vueuse\\shared',
npm WARN cleanup     [Error: EBUSY: resource busy or locked, rmdir 'C:\dev\vue-carbon-emit\node_modules\@vueuse\shared\node_modules\vue-demi'] {
npm WARN cleanup       errno: -4082,
npm WARN cleanup       code: 'EBUSY',
npm WARN cleanup       syscall: 'rmdir',
[...]

PR #1611, opened 2024-07-11
build: move dev postinstall

davidnixon:
Contributes to #1576 

## What did you do?

1. Move the local development postinstall (hooks for husky etc) out of the scripts area of package.json during packaging
2. Copy the package / prod postinstall (ibm telemetry) from "packrat" area of package.json to "scripts" during packaging

Since the dev postinstall is no longer in scripts there is no way for npm to find and run it. Right?
I think npm 10 may be running everything with a "_postinstall" suffix so another alternative would be to rename the "dev_postinstall"  to something else. Moving completely out of scripts area seems safer.
[carbon-vue-3.0.17-8eb61.tgz](https://github.com/user-attachments/files/16174354/carbon-vue-3.0.17-8eb61.tgz)

## Why did you do it?

node 20/npm 10 issues

## How have you tested it?

locally with local package tgz package

- [ ] N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,9 +7,11 @@
   "main": "dist/carbon-vue-3.umd.min.js",
   "types": "src/index.d.ts",
   "web-types": "dist/web-types.json",
+  "packrat": {
+    "prod_setup": "npx ibmtelemetry --config=telemetry.yml || true"
+  },
   "scripts": {
     "postinstall": "node devPreinstall.js && node .storybook/postinstall.js",
-    "pack_postinstall": "npx ibmtelemetry --config=telemetry.yml || true",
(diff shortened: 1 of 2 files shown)


davidnixon:
@guilherme-gm what do you think?

guilherme-gm:
wow that was fast :)
I think this might work, and I agree with you that having only one script with the `postinstall` word should be enough, unless npm is somehow copying from the repo (which I think it is unlikely). once you publish a release I can test it

guilherme-gm (approved):

gabisinhas:
Hello Guys , Can we reopen this ticket ?? I am facing the same issue even with the latest update from Carbon Vue 3.0.18
image

PR #1614, opened 2024-07-16
chore(ibm-common): adding mandatory intrumentation tags

jeffchew:
## What did you do?

Adding mandatory intrumentation tags for Carbon Vue storybook

## Why did you do it?

Per IBM OPRT audit, we are required to have proper instrumentation on all domains owned by IBM. This will bring the Carbon Vue storybook into compliance.

### DIFF
--- a/.storybook/manager-head.html
+++ b/.storybook/manager-head.html
@@ -74,6 +74,46 @@
   }, 1000);
 </script>
 
+<footer>
+  <dds-footer-container
+    key="footer"
+    disable-locale-button="true"
+    size="micro" />,
+</footer>
+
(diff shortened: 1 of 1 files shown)


davidnixon (approved):
LGTM but any chance the links can open in a new tab? I don't see an option for that in the docs for dds-footer but thought I would ask.

jeffchew:
> LGTM but any chance the links can open in a new tab? I don't see an option for that in the docs for dds-footer but thought I would ask.

Thank you @davidnixon ! Unfortunately the footer isn't set up with the ability to open those links to a new window.

PR #1617, opened 2024-07-25
fix: 1576 again

davidnixon:
Contributes to #1576 

## What did you do?
moved postinstall to its own js script

## Why did you do it?
fix npm 10+

## How have you tested it?
This time test both installing via a tgz file AND installing from a local registry.
Test steps:
- Confirm that installing 3.0.18 fails with npm 10 when installing (see reproduction steps in #1576)
- Confirmed that the same 3.0.18 also fails when installing with npm 10 from a local npm registry (see compose in local-dev and the updated README for how to test this.)
- Changed postinstall to a script that detects prod/dev install. 
- Built new version and posted this version to local npm registry
- Confirmed that the new 3.0.18+ installs properly with npm 10 from a local npm registry
- Confirmed that the dev install (`yarn install`) still installs the husky hooks & storybook custom controls

- [x] Yes


guilherme-gm (approved):
Tested on my machine with the local registry like you added in the README. Works for me too!

PR #1619, opened 2024-08-09
fix: dropdown menu fix for accessability announcement

davidnixon:
## What did you do?
Correct accessibility issue with dropdown menu.
Invalid text and warning text were not being announced in screen reader.

## Why did you do it?
Found during development by Alessio Mantegna. Fix also provided by Alessio.

## How have you tested it?
With screen reader in Safari.
Easiest place to test is with the slots story
http://localhost:6006/?path=/story/component-cvdropdown--slots
There you can toggle the warning and invalid slots and confirm they are announced.

- [x] N/A

### DIFF
--- a/src/components/CvDropdown/CvDropdown.vue
+++ b/src/components/CvDropdown/CvDropdown.vue
@@ -121,12 +121,22 @@
           </slot>
         </ul>
       </div>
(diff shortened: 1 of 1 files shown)


davidnixon:
Approved by Alessio, accessibility focal on my team.

Issue #1615, opened 2024-07-23
CvDataTable shows batch actions if table is empty


dovius:
## Detailed description

> Describe in detail the issue you're having.

CvDataTable displays batch actions even when the table is initially loaded empty. This behavior is unexpected as batch actions should not be visible unless there are items in the table.

> Is this a feature request (new component, new icon), a bug, or a general issue?

A bug.

> Is this issue related to a specific component?

Yes, it is related to the CvDataTable component.

> What did you expect to happen? What happened instead? What would you like to see changed?

I expected batch actions to be hidden when the table is empty on initial load. Instead, they are visible. I would like the batch actions to remain hidden until there are items in the table.

> What browser are you working in?

Chrome

> What version of the Carbon Design System are you using?

3.0.18

## Steps to reproduce the issue

1. Load the CvDataTable component with batch actions and an empty data set.
2. Observe that the header checkbox is marked and batch actions are visible.
3. Expect batch actions to be hidden when the table is initially empty.

## Additional information

The bug was introduced with this change: https://github.com/carbon-design-system/carbon-components-vue/commit/64cbe0c98c0d44375878eb8d478c6bafedd2215a#diff-4bd776d0344e699940e456c98121387ab65c6e4c80ba33960140e926ab309673L582

![Screenshot 2024-07-23 at 18 55 17](https://github.com/user-attachments/assets/ab089e50-b435-4ddf-8a39-a02aecdfcf1b)

igorgrv1:
Hi team, any update on this? Having the same issue

davidnixon:
Yep, definitely a bug and a regression. I added a workaround to the hello app. There the batch actions are added via the slot so, as a workaround, I tweaked the slot to add a " v-if=" like this:
```vue
          <template v-if="filteredFish.length > 0" #batch-actions>
            <cv-button :icon="hideIcon" @click="onHideSelected">{{
              t("hide")
            }}</cv-button>
          </template>
```
I'll queue up a fix asap

davidnixon:
@dovius @igorgrv1 
[carbon-vue-3.0.20-5c2d7.tgz](https://github.com/user-attachments/files/16912313/carbon-vue-3.0.20-5c2d7.tgz)

Give this one a try if you can
```sh
npm i carbon-vue-3.0.20-5c2d7.tgz
```

davidnixon:
Fixed in 3.0.21

PR #1623, opened 2024-08-25
fix: data table batch actions

davidnixon:
Contributes to #1615 

## What did you do?
Prevent batch actions from showing for empty tables

## Why did you do it?
Batch actions should not show for empty tables. This is a regression from 3.0.14 as noted in #1615 

## How have you tested it?
Implemented the table search function in the story book. Also tested in hello carbon vue app. 

- [x] Yes

### DIFF
--- a/src/components/CvCheckbox/CvCheckbox.stories.js
+++ b/src/components/CvCheckbox/CvCheckbox.stories.js
@@ -15,7 +15,7 @@ export default {
   argTypes: {
     ...storybookControlsFromProps(propsCvCheck),
     value: {
-      type: 'string',
+      type: { raw: 'string', required: true },
       description:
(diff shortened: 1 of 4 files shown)


davidnixon:
[carbon-vue-3.0.20-5c2d7.tgz](https://github.com/user-attachments/files/16741722/carbon-vue-3.0.20-5c2d7.tgz)
@dovius
@igorgrv1

lmk what you think of this fix

davidnixon:
[carbon-vue-3.0.20-5c2d7.tgz](https://github.com/user-attachments/files/16912304/carbon-vue-3.0.20-5c2d7.tgz)

davidnixon:
Just waiting on an independent test on this one

Issue #1625, opened 2024-09-11
Date Picker - type range


gabisinhas:
Date Picker type range is not returning event output with start date and end date. It is returning just start date.
![image](https://github.com/user-attachments/assets/043b495f-8c93-4d79-9d1b-5b25082e54a8)

## Detailed description

> Describe in detail the issue you're having.

> Is this a feature request (new component, new icon), a bug, or a general issue?
- bug

> Is this issue related to a specific component?
- Date Picker type range

> What version of the Carbon Design System are you using?
    "carbon-components": "^10.58.15",

## Additional information

- Screenshots or code
- Notes

davidnixon:
I will update the range version to emit the change event as an object similar to how the v-model works. 
I'll double check the react component to see how that version emits the change events.

davidnixon:
The onChange in Carbon react emits an array with 2 date objects. i.e.
```js
[Wed Sep 04 2024 00:00:00 GMT-0400 (Eastern Daylight Time), Thu Sep 19 2024 00:00:00 GMT-0400 (Eastern Daylight Time)]
```

The fix for this should follow that pattern.

davidnixon:
For `kind="single"` Carbon React issues an array of 1 date object. I was surprised by this as a array with a single object is not very intuitive but looking at the code, this event is coming directly from the underlying flat picker component. The fix for this issue should follow this same pattern of an array with single date object.

davidnixon:
fix in 3.0.21

PR #1626, opened 2024-09-18
fix: emit date change event to match React

davidnixon:
Contributes to #1625 

## What did you do?
- some minor eslint-style changes to the story to put `@` items at the end of the attribute list i.e. `@change="onChange"`
- some minor tidying of controls on stories
- add new emit for `dateChange` this event mimics the `change` event in the Carbon React
- separate the v-model stories for a `single` kind and `range` kind. The v-model is different for each of these and I think separating the stories makes it a bit clearer.
- for `single` kind and `range` kind stories, show the `dateChange` in the actions tab

## Why did you do it?

See #1625 but basically there was no way to listen for changes for a range style date picker.

## How have you tested it?

Via the storybook

- [x] Yes

### DIFF
--- a/src/components/CvDatePicker/CvDatePicker.stories.js
+++ b/src/components/CvDatePicker/CvDatePicker.stories.js
@@ -44,7 +44,7 @@ export default {
 
 const template = `
 <div>
-  <cv-date-picker v-bind="args" @change="onChange" :value="now">
+  <cv-date-picker v-bind="args" :value="now" @change="onChange">
   </cv-date-picker>
 </div>
 `;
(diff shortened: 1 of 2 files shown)


davidnixon:
The `change` event is unchanged. It still emits a native change event for the underlying input. This is used for kind `simple` and `short`.
image

The `change` event is still emitted for kind `single` and `range` but ti is not shown in the storybook.

davidnixon:
For kinds `single` and `range` the new `dateChange` event is emitted. The param passed to the event is an array of Date objects. For a kind `single` this array has 1 entry. For a kind `range`, the array has 2 entries.
image

davidnixon:
@gabisinhas
Can you have a look at this one and lmk. I did consider changing the `change` event but that would potentially break anyone who is currently using that event. That's why I decided on the `dateChange` event.

gabisinhas:
@davidnixon i think it might resolve this. If you can publish then I can test it. thanks for the attention to this issue.

davidnixon:
[carbon-vue-3.0.20-1abc8.tgz](https://github.com/user-attachments/files/17050403/carbon-vue-3.0.20-1abc8.tgz)

```sh
npm remove @carbon/vue
npm add carbon-vue-3.0.20-1abc8.tgz
```

Issue #1663, opened 2024-10-21
File upload component ignores `multiple=false` when dragging files


themoenen:
## Detailed description

**Describe in detail the issue you're having.**
The `multiple` flag for the [file uploader compoment](https://vue.carbondesignsystem.com/?path=/story/component-cvfileuploader--default-story) (which controls whether you can upload more than one file or not) is ignored when you drag files onto it.

**Is this a feature request (new component, new icon), a bug, or a general issue?**
Bug

**Is this issue related to a specific component?**
The [file uploader compoment](https://vue.carbondesignsystem.com/?path=/story/component-cvfileuploader--default-story)

**What did you expect to happen? What happened instead? What would you like to see changed?**
It should be impossible to select more than one file when `multiple` is set to `false`, regardless of the way the files are added.

**What browser are you working in?**
Chrome

**What version of the Carbon Design System are you using?**
3.0.21

**What offering/product do you work on? Any pressing ship or release dates we should be aware of?**
[OpenAD](https://github.com/acceleratedscience/open-ad-toolkit/) - No urgency as we'll build our own alternative component.

```
<cv-file-uploader
  label="Dataset"
  accept=".csv"
  helperText="Upload your dataset as CSV file."
  :multiple="false"
  @change="(e: Event) => emit('change', e)"
/>
```

![Image](https://github.com/user-attachments/assets/968909c1-221b-41b1-90b7-aa2e50983627)

themoenen:
No request for additional info was made. Please let me know if you need any additional information.

davidnixon:
this has been fixed. See PR merged above.

PR #1693, opened 2024-12-23
fix: adjust CvFileUploader drop handler to not ignore 'multiple' attribute

felipebritor:
Contributes to #1663 

## What did you do?

Update 'drop' event handler, at CvFileUploader, to handle files considering 'multiple' attribute

## Why did you do it?

Fix the drop behavior when multiple is falsy

## How have you tested it?

Yes. Automated tests also added.

- [ ] N/A
- [X] No
- [ ] Yes

### DIFF
--- a/src/components/CvFileUploader/CvFileUploader.stories.mdx
+++ b/src/components/CvFileUploader/CvFileUploader.stories.mdx
@@ -112,9 +112,10 @@ export const argTypes = {
   // attributes
   multiple: {
     type: 'boolean',
+    defaultValue: true,
     table: {
       type: { summary: 'boolean' },
-      category: 'attributes',
+      category: 'props',
     },
     description: 'standard input attribute for file type, allow multiple files to be added in a single prompt',
   },
--- a/src/components/CvFileUploader/CvFileUploader.vue
+++ b/src/components/CvFileUploader/CvFileUploader.vue
@@ -97,7 +97,7 @@
 </template>
 
 <script setup>
-import { ref, computed, onMounted, watch } from 'vue';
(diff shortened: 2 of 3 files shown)


davidnixon:
@felipebritor a tweak is still needed. 
1. This does prevent me from selecting multiple files and dropping them on the target. When I do that I only get 1 file as expected. ✅ 
2. I can however drag and drop a file and then drag and drop another file so that now there are two! ❌ 
3. The multiple attribute should default to true ❌ 
4.  The React story is unclear on how the button mode (non drag & drop mode) should work. I wonder if you or @themoenen has an opinion on that?

Referencing the same component in res=act:
https://react.carbondesignsystem.com/?path=/story/components-fileuploader--drag-and-drop-upload-container-example-application&args=multiple:!false

felipebritor:
> @felipebritor a tweak is still needed.
> 
> 1. This does prevent me from selecting multiple files and dropping them on the target. When I do that I only get 1 file as expected. ✅
> 2. I can however drag and drop a file and then drag and drop another file so that now there are two! ❌
> 3. The multiple attribute should default to true ❌
> 4. The React story is unclear on how the button mode (non drag & drop mode) should work. I wonder if you or @themoenen has an opinion on that?
> 
> Referencing the same component in res=act: https://react.carbondesignsystem.com/?path=/story/components-fileuploader--drag-and-drop-upload-container-example-application&args=multiple:!false

Regarding your points:
2. Sorry about it, David. Fixed
3. Added to PR. Had to turn multiple into a props for it
4. I think it does [here](https://react.carbondesignsystem.com/?path=/story/components-fileuploader--default). Or did you mean something else?

PR #1701, opened 2025-01-16
fix: make ids consistent

davidnixon:
Contributes to #

## What did you do?
Make id assignments consistent. For instance, in the data table an id can be assigned
```vue
<cv-data-table id="my-table" .../>
```
But the data table includes a pagination component but the data table component user cannot set the id for this. This leads to a nuxt hydration warning.

The fix is to use the id assigned to the table to set the id of the pagination component.
```vue
 <cv-pagination :id="`pagination-${uid}`" .../>
```

## Why did you do it?
Nuxt hydration complains about issues like this. The ds and other attributes should be consistent between the server and client but when the ids are randomly generated it causes a warning.

## How have you tested it?

Local nuxt project

- [ ] N/A

### DIFF
--- a/src/components/CvDataTable/CvDataTable.vue
+++ b/src/components/CvDataTable/CvDataTable.vue
@@ -160,6 +160,7 @@
                   <cv-checkbox-skeleton v-if="isSkeleton" />
                   <cv-checkbox
                     v-else
+                    :id="`col-cb-${uid}`"
                     v-model="headingChecked"
                     :form-item="false"
                     value="headingCheck"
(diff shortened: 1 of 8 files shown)


davidnixon:
I need to do a bit more cleanup here. I see some inconsistencies with the way the id is assign. Sometimes it is `${id}-suffix` and sometimes `prefix-${id}`. Sometimes its `${uid}` etc.

davidnixon:
@felipebritor could you have a look at this one.

src/components/CvDataTable/CvDataTable.vue
```diff
@@ -73,20 +73,20 @@
               <Search16 :class="`${carbonPrefix}--search-magnifier-icon`" />
             </div>
             <label
-              :id="`${uid}-search`"
-              :for="`searchbox-${uid}`"
+              :id="`${cvId}-search`"
```
davidnixon, line 76: This one was already here using "search" as a prefix so I do not want to change to a suffix.

felipebritor:
@davidnixon sure, I'll check tonight

felipebritor (approved):
Sorry for the delay, @davidnixon. LGTM

Issue #1714, opened 2025-02-26
CvDataTable index.js not exporting all its related named components


oskarnrk:
CvDataTable index.js exports only CvDataTable component but doesn't export some other named components used inside it, e.g. CvDataTableRow. 

## Steps to reproduce the issue

`import { CvDataTableRow } from "@carbon/vue";`

davidnixon:
I cannot reproduce this one. Do you have a repo or a small sample?
All the components are exported from https://github.com/carbon-design-system/carbon-components-vue/blob/main/src/index.js

I do agree that the CvDataTable index.js format does not match the other index.js files though.

oskarnrk:
@davidnixon sorry, you're right, I did a mistake reporting the import in "Steps to reproduce the issue", I imported `@carbon/vue/src/components/CvDataTable`, using just `@carbon/vue` it works. So, we can say that is just to match the other index.js format.

PR #1715, opened 2025-02-26
fix: cv-data-table related named components export

oskarnrk:
Contributes to #1714 

## What did you do?
Just exported the CvDataTable related named components

## How have you tested it?
Changing the index.js file in node_modules after install. 

- [ ] N/A
- [X] No
- [ ] Yes


davidnixon:
Added comment to issue

davidnixon (approved):
LGTM

Issue #1710, opened 2025-02-13
Number input not working in FF and Safari


sfruzsi:
## Number input accepting non-number values

-->

## Detailed description

> When using **cv-number-input** the component accepts non-number values in different browsers (Firefox and Safari) while in Chrome it works as intended and only accepts number values.

## Steps to reproduce the issue

1. Open https://vue.carbondesignsystem.com/?path=/story/component-cvnumberinput--default in Firefox
2. Write letters into the number input field.

## Additional information

![Image](https://github.com/user-attachments/assets/1f7fbb50-a3b9-4a3b-9269-ba72868e263e)

sfruzsi:
@davidnixon Hi, this is what we discussed on Slack, hope I opened the issue in the correct place. Thanks!

davidnixon:
@sfruzsi can you test this PR? [fix: prevent non-numeric input on firefox and Safari #1716](https://github.com/carbon-design-system/carbon-components-vue/pull/1716)

davidnixon:
PR deployed here:
https://davidnixon.github.io/carbon-compPR deployed here:
https://davidnixon.github.io/carbon-components-vue/?path=/story/component-cvnumberinput--defaultonents-vue/?path=/story/component-cvnumberinput--default

sfruzsi:
Hi @davidnixon , it works now in Safari and Firefox. Thank you for the fix!

PR #1716, opened 2025-02-28
fix: prevent non-numeric input on firefox and Safari

davidnixon:
Contributes to #1710 

## What did you do?

 prevent non-numeric input on firefox and Safari

## Why did you do it?

Match React component where similar issue does not exist.

## How have you tested it?

Storybook in FF and Safari

- [ ] N/A

### DIFF
--- a/src/components/CvNumberInput/CvNumberInput.vue
+++ b/src/components/CvNumberInput/CvNumberInput.vue
@@ -98,7 +98,7 @@
 </template>
 
 <script setup>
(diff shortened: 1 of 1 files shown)


davidnixon:
PR deployed here:
https://davidnixon.github.io/carbon-components-vue/?path=/story/component-cvnumberinput--default

davidnixon:
Approved in the issue and slack

Issue #1724, opened 2025-05-06
Issue with inline svg in _CvSvg.vue


rusetski:
Hi there!

Inline SVG in Vue 3 is not working now due to props format changes in Vue 3, in Vue 3 no longer needs domProps.

The problem is in this part of the code https://github.com/carbon-design-system/carbon-components-vue/blob/c16922d8aa89ae037f3ff0d59e1b6f9a4bb71cd1/src/components/CvSvg/_CvSvg.vue#L36

Link to the Vue documentation: https://v3-migration.vuejs.org/breaking-changes/render-function-api#vnode-props-format

rusetski:
Here is PR: https://github.com/carbon-design-system/carbon-components-vue/pull/1725

PR #1725, opened 2025-05-06
fix: for CvSvg, update props structure for Vue 3

rusetski:
Fix CvSvg not rendering inline svg properly (#1724)

### DIFF
--- a/src/components/CvSvg/_CvSvg.vue
+++ b/src/components/CvSvg/_CvSvg.vue
@@ -33,7 +33,7 @@ export default {
         return h(props.svg);
       } else if (isSvg.value) {
         // an SVG document
-        return h('svg', { domProps: { innerHTML: props.svg } });
+        return h('svg', { innerHTML: props.svg });
       } else {
         if (isSymbol.value) {
           return h('svg', {}, [h('use', { attrs: { href: props.svg } })]);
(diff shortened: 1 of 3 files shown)


guilherme-gm:
Thank you for your contribution, overall it looks good to me.

I noticed we have 1 test that was supposed to detect this issue, but it is doing an assertion that is too basic to get it.

May you update for us?

On `src/components/CvSvg/__tests__/CvSvg.spec.js`, we have this:

```JS
  it('renders the SVG inline', () => {
    const wrapper = shallowMount(CvSvg, {
      props: { svg: svgInline },
    });

    expect(wrapper.find('svg')).not.toBeNull();
  });
```

My suggestion is to change this assertion to a snapshot of the HTML, like that:

```JS
  it('renders the SVG inline', () => {
    const wrapper = shallowMount(CvSvg, {
      props: { svg: svgInline },
    });

    // *** Change here ***
    expect(wrapper.find('svg').html({ raw: true })).toMatchSnapshot();
  });
```

so we can actually see the Svg being rendered in case something breaks again in the future.

In case you are not used to snapshots: after your change this test, just run `npm run test` afterwards so it generates the snapshot file and commit it too.

Thank you!

rusetski:
Hi, @guilherme-gm! I added a new commit with your changes

guilherme-gm (approved):
Thank you for updating it! Approved

felipebritor (approved):

PR #1731, opened 2025-05-21
fix: reset filter value when prop options change

pac0rro:
Contributes to no issue

## What did you do?
Reset the filter value when prop options change

## Why did you do it?
Because when the options array change the previous filter usually doesn't match the new options

## How have you tested it?
In local with Chrome and Firefox

There was no need

- [  ] N/A

### DIFF
--- a/src/components/CvComboBox/CvComboBox.vue
+++ b/src/components/CvComboBox/CvComboBox.vue
@@ -278,6 +278,7 @@ watch(
 watch(
   () => props.options,
   () => {
+    filter.value = '';
     updateOptions();
   }
 );


felipebritor:
Hello, @pac0rro. Thanks for your contribution.

The change is good, but I couldn't find anything in the [carbon guildelines](https://carbondesignsystem.com/components/dropdown/usage/#combo-box) to back it up as a combox behavior. Maybe we could have it enabled by a prop? Let me know what you think.

pac0rro:
Hi, Felipe.

Probably you are right. But there are other behaviors in this component that doesn't match the guidelines or the reference implementation (React), for example the click on the filter text button.

However, I am fine with a new property to enable this operation. Do you want me to add this property ... or will you do it instead?

pac0rro:
Hi, Felipe.

Probably you are right. But there are other behaviors in this component that doesn't match the guidelines or the reference implementation (React), for example the click on the filter text button.

However, I am fine with a new property to enable this operation. Do you want me to add this property ... or will you do it instead?

pac0rro:
Sorry, Felipe, I closed this PR by accident.

Ok, I will update this request when the property is created. This will take more work because I would need to modify the docs, story, etc.

I will create a prop with name "resetFilterOnOptionsChange". Do you agree?

PR #1786, opened 2025-10-17
feat: fix CvUiShell accessibility problems

Marghen:
## What did you do?

I fixed the accessibility problems that even Storybook was pointing out for the CvUiShell component.

## Why did you do it?

Because we use it in our project and it needs to be 100% accessibility compliant.

## How have you tested it?

With StoryBook accessibility testing.

- [X] N/A
- [ ] No
- [ ] Yes


Marghen:
Hey @felipebritor, any news perhaps? Thank you

PR #1811, opened 2026-03-19
fix: cvmodal closing issue on mouse up

TagajN:
Contributes to #

## What did you do?
I added mouse down event to follow and check which element was clicked

## Why did you do it?
Because in Chrome when user click on modal and release that click outside of it, then modal is closing.

## How have you tested it?
User test

- [ ] N/A
- [ ] No
- [ ] Yes


guilherme-gm (approved):

guilherme-gm:
Thanks for the contribution!

PR #1816, opened 2026-03-31
fix(dropdown): rerender with same value

99RP:
Contributes to #

Hello, it is my first time contributing to your project, please let me know if there is anything that i could be doing differently or have made any mistakes. Thanks in advance.

## What did you do?
Added a guard to the `modelValue` watcher in `CvDropdown` so it skips updating when the new value is identical to the one already selected.

## Why did you do it?
The watcher previously cleared the internal caption on every prop update, even when the value hadn't changed. This caused the displayed label to revert to the placeholder text whenever a parent component rerendered with the same modelValue

I found this out because my application does optimistic frontend update before receiving the true updated data from the backend. In my case backend returned the same value that was already optimistically updated and the dropdown displayed the placeholder.

## How have you tested it?
Added two tests, they failed before and now they pass. Also tested in my real world application and it works as expected now.

- [x] N/A
- [ ] No
- [ ] Yes

### DIFF
--- a/src/components/CvDropdown/CvDropdown.vue
+++ b/src/components/CvDropdown/CvDropdown.vue
@@ -296,6 +296,7 @@ const emit = defineEmits(['update:modelValue', 'change']);
 watch(
   () => props.modelValue,
   () => {
+    if (props.modelValue === dataValue.value) return;
     dataValue.value = props.modelValue;
     dataCaption.value = undefined;
   }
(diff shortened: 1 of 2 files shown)


guilherme-gm (approved):

guilherme-gm:
sorry for the delay; thank you for the contribution!

PR #1842, opened 2026-08-13
fix(CvPagination): stop stale change re-emits, add opt-in controlled mode and i18n formatters

Marghen:
Description:

## What did you do?

Fixed CvPagination re-emitting a change event with stale data when numberOfItems changes by itself. Also added an opt-in controlled mode with **v-model:page** and **v-model:page-size**, plus optional rangeTextFormatter and **pageOfPagesFormatter** props for i18n.

Added a few stories to cover the new controlled mode and formatters.

## Why did you do it?

The numberOfItems watcher was triggering **change** as if the user had picked a new page size, even when the data was simply updated.

Also, page and pageSize were only used as initial values. This meant a parent couldn't really reject a page change (the component could silently get out of sync and eventually get stuck)

The new formatter props make it possible to translate the range/page text without having to reimplement the slots.

## How have you tested it?

Added tests covering all three changes  (12/12 passing)

The full suite is still green (70 suites / 502 tests) with lint + build + build-web-types all passing.

Also did a quick manual check in Chrome via Storybook to make sure the old page-size-reset bug (the one the removed nextTick workaround was guarding against) is actually gone, and everything works.

- [ ] N/A
- [ ] No
- [x] Yes

### DIFF
--- a/src/components/CvPagination/CvPagination.stories.mdx
+++ b/src/components/CvPagination/CvPagination.stories.mdx
@@ -3,9 +3,29 @@ import { CvPagination } from '.';
 import { sbCompPrefix } from '../../global/storybook-utils';
 import { action } from '@storybook/addon-actions';
 import { ref } from 'vue';
-const myPage=ref(1);
-const myPageSizes=ref([5,7,10,11,13])
-const myNumItems=ref(13)
+const myPage = ref(1);
+const myPageSizes = ref([5, 7, 10, 11, 13]);
+const myNumItems = ref(13);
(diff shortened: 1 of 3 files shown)


Marghen:
@felipebritor Hi Felipe, could you have a look when you have time? Thank you!

felipebritor:
Hello, @Marghen . Thanks for the PR. I'll check it on friday/weekend or earlier if possible

guilherme-gm (approved):
Thank you for the contribution!

